PR #54, opened 2021-06-01
Add duplicate id check for existing package identifiers

### LINKED ISSUE #30: `wingetcreate update` should not be case-sensi-weird

I was trying to update node.js to the latest ver of of the 14.x branch, but by running `winget show openjs.nodejs --versions` I noticed its missing 14.16.1 and 14.17.0.

So I tried submitting it:

```sh
PS C:\Users\mnpen> wingetcreate new https://nodejs.org/dist/v14.17.0/node-v14.17.0-x64.msi
Downloading and parsing: https://nodejs.org/dist/v14.17.0/node-v14.17.0-x64.msi...

This tool will walk you through a series of questions to help you create your package manifest.

For information about the restrictions for each field, visit https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md

Press ENTER to submit the value for each question including accepting the (default) value.

Please enter values for the following fields:
 [PackageIdentifier] The package unique identifier |e.g. Microsoft.VisualStudio|: OpenJS.Nodejs
 [PackageVersion] The package version |e.g. 1.2.3.4|: 14.17.0
 [DefaultLocale] The package meta-data default locale |e.g. en-US|: en-US
 [Publisher] The publisher name |e.g. Microsoft|: Node.js Foundation
 [PackageName] The package name |e.g. Visual Studio|: Node.js
 [License] The package license |e.g. MIT License|:
```

OK... it doesn't seem to be recognizing that this is an existing package. Maybe there's a different command for updating the existing package? I cancel out, run ` wingetcreate --help`, ahah, yes, of course there is, it's called `update`. 

```
PS C:\Users\mnpen> wingetcreate update -i OpenJS.Nodejs -u https://nodejs.org/dist/v14.17.0/node-v14.17.0-x64.msi
A GitHub account or personal access token must be linked in order to continue with this command.
Execution of the command is paused until the GitHub login has been completed.

Initiating GitHub login...

Launching web browser... If default web browser does not launch, please go to: https://github.com/login/device
Please enter the following user verification code when prompted: <bleep>

Resuming command execution...
Retrieving latest manifest for OpenJS.Nodejs
ERROR: repos/microsoft/winget-pkgs/contents/manifests/o/OpenJS/Nodejs was not found.
Please check the spelling and casing of the package identifier
```

Well that's really weird because when I run `winget list` it explicitly says I have `OpenJS.Nodejs` installed:

```
Node.js                                  OpenJS.Nodejs                                        14.16.1             16.2.0       winget
```

So I go to github to try and figure out what it's *actually* listed as. I find `https://github.com/microsoft/winget-pkgs/tree/master/manifests/o/OpenJS/NodeJS` and try again, changing it to capital `JS` to match the github URL. And indeed, that seems to have found something:

```
PS C:\Users\mnpen> wingetcreate update -i OpenJS.NodeJS -u https://nodejs.org/dist/v14.17.0/node-v14.17.0-x64.msi
Using GitHub token from cache...
Retrieving latest manifest for OpenJS.NodeJS
Downloading and parsing: https://nodejs.org/dist/v14.17.0/node-v14.17.0-x64.msi...

Generating a preview of your manifests...
Version manifest preview:
# Created using wingetcreate

PackageIdentifier: OpenJS.NodeJS
PackageVersion: 16.2.0
DefaultLocale: en-US
ManifestVersion: 1.0.0
...
```

But it thinks its 16.2.0 for some reason and doesn't even prompt to ask otherwise. But maybe that's issue #23 ?

Anyway, I'm calling this sensi-weird because it *does* appear to be case-sensitive but it *doesn't* match the case that `winget list` shows, so that's kinda weird.

# Steps to reproduce

Try running `wingetcreate update -i` using an existing package but "wrong" case.

# Expected behavior

Just do a case-insensitive match. If different packages are allowed to have the same name but in different cases, that's going to be a whole 'nother problem.

And also, maybe a separate issue, but make `winget list` show the actual ID in the ID column.

# Actual behavior

Doesn't match

# Environment

```none
Windows Manifest Creator v0.1.0.1
```

### LINKED ISSUE #47: `wingetcreate new` ignores already existing folder structure.
`wingetcreate new` ignores the existing folder structure and can cause major issues.

# Steps to reproduce
https://github.com/microsoft/winget-pkgs/pull/14367
`wingetcreate new voidtools.everything url`

# Actual behavior
Creates the manifests in `manifests/v/voidtools/everything/1.4.1.1005`

# Expected behavior
To create the manifest in `manifests/v/voidtools/Everything/1.4.1.1005`

# Environment
Windows Manifest Creator v0.1.0.1

```none
Windows Package Manager Manifest Creator version? (wingetcreate.exe -?)

Any other software?
```

ryfu-msft:
Fixes #47
Resolves #30 

To partially address the recent influx of duplicate manifests being created, this pull request adds an additional duplicate id check to inform the user if the package id they are using to create a `new` manifest already exists in the winget-pkgs repository. The logic is as follows:

- After a user confirms their packageID, wingetcreate will search winget-pkgs to determine if that directory path already exists.
- PackageId check is not case sensitive, so even if the packageId differs by capitalization, the check will still pick that up. 
- If the packageId already exists, an error will be shown and the tool will terminate without publishing the manifests.
- Update command is no longer case sensitive and utilizes the same duplicate check function to obtain the exact packageId for you.

Other changes:
- Added the version of the tool to the header of the manifest along with the program name.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/54)


src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -200,6 +200,17 @@ public async Task<string> GetFileContentsAsync(string path)
...
         {
             await this.github.Repository.Get(this.wingetRepoOwner, this.wingetRepo);
+        }
+
+        /// <summary>
+        /// Recursively searches the repository for the provided package identifer to determine if it already exists.
+        /// </summary>
+        /// <param name="packageId">The package identifier.</param>
+        /// <returns>Boolean value indicating if a duplicate manifest exists.</returns>
+        public async Task<bool> CheckDuplicatePackageId(string packageId)
```
palenshus, line 210: >CheckDuplicatePackageId [](http://example.com/codeflow?start=32&length=23)

This will return true even if the packageId is only a substring of something in the repo, for example, microsoft.visualstudio.2019 returns true.

I think this is probably fine, as I can't imagine we'd want to allow a package in that's just a substring of another one, but @denelon to confirm

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -138,10 +139,18 @@ public override async Task<bool> Execute()
...
                     this.OutputDir = Directory.GetCurrentDirectory();
+                }
+
+                GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+                if (await client.CheckDuplicatePackageId(manifests.VersionManifest.PackageIdentifier))
+                {
+                    Console.WriteLine();
+                    Logger.ErrorLocalized(nameof(Resources.PackageIdAlreadyExists_Error));
+                    return false;
                 }
```
palenshus, line 150: Should we move this to be right after line 131?

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -226,6 +237,28 @@ private static string GetJwtToken(string gitHubAppPrivateKeyPem, int gitHubAppId
...
+            if (index == packageId.Length)
+            {
+                return true;
+            }
+
+            var contents = await this.github.Repository.Content.GetAllContents(this.wingetRepoOwner, this.wingetRepo, path);
+
+            foreach (RepositoryContent content in contents)
+            {
+                if (string.Equals(packageId[index].ToLowerInvariant(), content.Name.ToLowerInvariant()))
```
palenshus, line 253: >packageId[index].ToLowerInvariant() [](http://example.com/codeflow?start=34&length=35)

you can make this a variable above the loop so you only calculate it once

palenshus (changes requested):
:clock1:

src/WingetCreateCore/Common/GitHub.cs
denelon: Take a look at Visual Studio Code, and Visual Studio Code Insiders.

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: Changed to suggestion

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: In my opinion, I think this is okay. Even if the inputted packageId is only a substring, that still is something we shouldn't allow because it is matching an existing directory. We want to ensure that the new packageId the user is creating is completely unique.

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: I thought about this earlier and I decided not to put it there because we still give the user an opportunity to go back and change their values and they could change it correctly the second time. If we were to put it right after line 131, that would mean the duplicate check only takes place after all of the fields from the VersionManifest have been prompted, which would end on default locale. This would seem odd to the user that the check happens after default locale and not packageId. This is just my thinking so if you feel strongly about it, I'd be okay with changing it.

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -200,6 +200,18 @@ public async Task<string> GetFileContentsAsync(string path)
...
             await this.github.Repository.Get(this.wingetRepoOwner, this.wingetRepo);
+        }
+
+        /// <summary>
+        /// Recursively searches the repository for the provided package identifer to determine if it already exists.
+        /// </summary>
+        /// <param name="packageId">The package identifier.</param>
+        /// <returns>Tuple value with T1 being a boolean that represents whether the inputted packageId matches an existing one
+        /// and T2 as the string of the matching, (case-sensitive) exact packageId. </returns>
+        public async Task<(bool IsMatch, string ExactId)> CheckDuplicatePackageId(string packageId)
```
palenshus, line 211: >CheckDuplicatePackageId [](http://example.com/codeflow?start=58&length=23)

Why not just call this FindPackageId or something and have it return the case-sensitive match if found, otherwise null

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -95,12 +95,20 @@ public override async Task<bool> Execute()
...
             {
-                Logger.DebugLocalized(nameof(Resources.RetrievingManifest_Message), this.Id);
+                Logger.DebugLocalized(nameof(Resources.RetrievingManifest_Message), this.Id);
+
+                GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+                var (isMatch, exactId) = await client.CheckDuplicatePackageId(this.Id);
+
+                if (isMatch)
+                {
+                    this.Id = exactId;
```
palenshus, line 105: Nice! With this I think you can link this as a fix for #30 also

palenshus: Ah, I see you're ahead of me :)

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Talked about this offline. We agreed that prompting package ID should be separate from the other fields as this requires an additional duplicate check.

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: I agree. Changed to suggestion. FindPackageId now returns the match if found, null otherwise.

palenshus (approved):
:shipit:

PR #55, opened 2021-06-02
Add check for handling updating multiple distinct installer url manifests

### LINKED ISSUE #15: Update command: Bad behavior if existing manifest has multiple installer URLs and a URL is provided on cmd-line

For the Update command, if the existing manifest in winget-pkgs has multiple installer urls, we'll get bad behavior if the URL doesn't specify a URL on the command line.

# Steps to reproduce

update -i Microsoft.PowerShell -u https://github.com/PowerShell/PowerShell/releases/download/v7.1.3/PowerShell-7.1.3-win-x64.msi

# Expected behavior

Error message indicating that multiple URLs are not supported.

# Actual behavior

Update "succeeds" and produces an installer manifest with two nodes with the same URL, but different archs and hashes

ryfu-msft:
This pull request addresses the multiple installer url bug fixes #15 

Changes:
- Checks the manifest to be updated for multiple distinct installer urls and notifies the user that updating is currently not supported for this scenario.
- Changes the --id parameter to now be positional instead of a flag.
- Refactored Update Command unit tests
- Refactored Update Command 

Testing:
- Added a test that utilizes a manifest with multiple distinct urls to ensure that this is verifies. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/55)


src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -1,79 +1,106 @@
...
+
+        /// <summary>
+        /// Tests the update command to ensure that updating a multiple installer manifests should throw an error.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task UpdateWithMultipleInstallers()
+        {
+            UpdateCommand command = this.GetUpdateCommand(TestConstants.TestMultipleInstallerPackageIdentifier, null, this.tempPath);
+            Assert.IsFalse(await command.Execute(), "Command should have failed");
```
palenshus, line 88: >Execute [](http://example.com/codeflow?start=41&length=7)

Isn't the point of refactoring UpdateCommand.cs so that you can pass manifest yamls directly to it from test cases, instead of requiring new files in winget-pkgs-submission-test?

palenshus (changes requested):
:clock1:

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: You are right, I wasn't quite sure if we wanted to completely abandon the entire update command/github flow, but I see your point now. I have made the change so that we can pass local manifests directly to the tests

src/WingetCreateTests/WingetCreateTests/TestUtils.cs
```diff
@@ -55,6 +56,16 @@ public static void SetMockHttpResponseContent(string installerName)
...
+        }
+
+        /// <summary>
+        /// Obtains the relative filepath of the resources test data directory.
+        /// </summary>
+        /// <param name="fileName">File name of the test file.</param>
+        /// <returns>Full path of the test file.</returns>
+        public static string GetTestFile(string fileName)
+        {
+            return Path.Combine(Environment.CurrentDirectory, @"Resources\", fileName);
```
palenshus, line 68: >\ [](http://example.com/codeflow?start=73&length=1)

no biggie but for future reference you don't need the trailing slash here

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -1,79 +1,118 @@
...
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task UpdateVersion()
+        {
+            string version = "1.2.3.4";
+            UpdateCommand command = this.GetUpdateCommand(TestConstants.TestPackageIdentifier, version, this.tempPath);
+            SingletonManifest testManifest = Serialization.DeserializeFromPath<SingletonManifest>(TestUtils.GetTestFile($"{TestConstants.TestPackageIdentifier}.yaml"));
+            List<string> manifests = new List<string> { testManifest.ToYaml(UpdateCommandTestHeader) };
+            manifests.Add(testManifest.ToYaml(UpdateCommandTestHeader));
```
palenshus, line 63: Why are you adding this file twice?

palenshus, line 63: Unless you're wanting to explicitly test the yaml round-tripping, you could just change this to a new List<string> { File.Read(...yaml) }

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -1,79 +1,118 @@
...
+    using Microsoft.WingetCreateCore;
+    using Microsoft.WingetCreateCore.Common;
+    using Microsoft.WingetCreateCore.Models.Singleton;
+    using Microsoft.WingetCreateTests;
+    using NUnit.Framework;
+
+    /// <summary>
+    /// Test cases for verifying that the "update" command is working as expected.
+    /// </summary>
+    public class UpdateCommandTests : GitHubTestsBase
```
palenshus, line 23: >GitHubTestsBase [](http://example.com/codeflow?start=38&length=15)

Correct me if I'm wrong, but I believe you can now remove this base class since this test doesn't have to hit github anymore (cool!)

src/WingetCreateTests/WingetCreateTests/TestUtils.cs
ryfu-msft: i knew that 😅. Removed extra trailing slash.

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: Overlooked this, removed this

ryfu-msft: I think I agree that testing the yaml round-tripping is a bit excessive here. Changed to suggestion

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: Talked offline, I moved the InvalidPackageId test to the GitHub unit tests as it is not needed for testing the update command. Also removed the base class from UpdateCommandTests and NewCommandTests

palenshus (approved):
:shipit:

PR #56, opened 2021-06-02
Clean up temp files from installer download

### LINKED ISSUE #40: Installer hash does not match
If creating a manifest file, and then needing to update it, due to a new version getting released, you need to remember to delete the original file from the temp folder.
I double checked several times, the file in temp IS updated to the new version, yet if I do not delete it before the second wingetcreate it fails when you validate the manifest using winget install.
I have tested this in 2 different machines with same result.
# Steps to reproduce
```
wingetcreate new https://github.com/open-rpa/openrpa/releases/download/1.2.47/OpenRPA.msi
wingetcreate new https://github.com/open-rpa/openrpa/releases/download/1.2.48/OpenRPA.msi
winget install --manifest manifests\o\OpenRPAApS\OpenRPA\1.2.48.0
```
if you how ever delete the file from temp, it works
```
wingetcreate new https://github.com/open-rpa/openrpa/releases/download/1.2.47/OpenRPA.msi
del %temp%\openrpa.msi
wingetcreate new https://github.com/open-rpa/openrpa/releases/download/1.2.48/OpenRPA.msi
winget install --manifest manifests\o\OpenRPAApS\OpenRPA\1.2.48.0
```
# Expected behavior
Not needing to delete the file in temp
# Actual behavior

# Environment
Windows Manifest Creator v0.1.0.1

ryfu-msft:
The changes in this pull request ensure that the temp files from the downloaded installer are deleted before the end of the command execution. This should fix #40, where the installer hash does not match due to old installer files not being cleaned up from the %TEMP% directory.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/56)


palenshus:
Looks like we have a bug, in DownloadFileAsync, if the file already exists, but with a different length, we download the new one, but looks like that doesn't replace the old one, it just writes overwrites it. Meaning, if the previous one is larger, we'll download a smaller one and write those bytes over the old one, creating a bogus file.

The fix is simple, and I think sufficient to protect us here, add a File.Delete:

            if (!File.Exists(targetFile) || new FileInfo(targetFile).Length != downloadSize)
            {
                File.Delete(targetFile);
                using var targetFileStream = File.OpenWrite(targetFile);

What do you think?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -205,8 +207,9 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
 
                     var currentValue = property.GetValue(manifest);
                     var result = PromptProperty(manifest, currentValue, property.Name);
-                    property.SetValue(manifest, result);
-                    Logger.Trace($"Property [{property.Name}] set to the value [{result}]");
+                    property.SetValue(manifest, result);
+                    string formattedResult = result.ToString().Replace("{", string.Empty).Replace("}", string.Empty);
+                    Logger.Trace($"Property [{property.Name}] set to the value [{formattedResult}]");
                 }
```
palenshus, line 213: I don't think this addresses their issues. All the stack traces seem to originate in this call in Program.cs:

    Logger.Error(ex.ToString());

I believe that what's possibly happening is that the exception method has a format string inside of it, for example "{version}". You can repro with:

wingetcreate new [https://foo.com/{version}.exe](https://foo.com/{version}.exe)

palenshus: We should fix this in a separate PR

ryfu-msft:
> Looks like we have a bug, in DownloadFileAsync, if the file already exists, but with a different length, we download the new one, but looks like that doesn't replace the old one, it just writes overwrites it. Meaning, if the previous one is larger, we'll download a smaller one and write those bytes over the old one, creating a bogus file.
> 
> The fix is simple, and I think sufficient to protect us here, add a File.Delete:
> 
> ```
>         if (!File.Exists(targetFile) || new FileInfo(targetFile).Length != downloadSize)
>         {
>             File.Delete(targetFile);
>             using var targetFileStream = File.OpenWrite(targetFile);
> ```
> 
> What do you think?

100% agree. This is a much better solution. I added your suggested changes and removed mine

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: I reverted my changes because your recent PR fixes this issue.

palenshus: (Fixed in separate PR)

NOTICE.txt
```diff
@@ -369,18 +369,17 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
...
-(c) 2008 VeriSign, Inc.
-(c) Microsoft Corporation.
-
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+---------------------------------------------------------
+
+---------------------------------------------------------
+
+jose-jwt 3.1.1 - MIT
```
palenshus, line 378: >jose [](http://example.com/codeflow?start=0&length=4)

Thanks!

palenshus:
>     using CommandLine;

Can you reset this file since it's no longer being changed?

---
Refers to: src/WingetCreateCLI/Commands/NewCommand.cs:13 in 9c904ab. [](commit_id = 9c904abf36320082d493027a0a49762dfccccb27, deletion_comment = False)

palenshus (approved):
:shipit:

PR #91, opened 2021-06-23
Removing BOM prefix from manifest content

### LINKED ISSUE #90: YamlDotNet.Core.YamlException: (Line: 1, Col: 1, Idx: 0)
# Brief description of your issue /  Steps to reproduce /  Actual behavior

```
PS C:\Users\******\GitHub\winget-pkgs> wingetcreate update Balena.BalenaCLI -v 12.44.21 -u https://github.com/balena-io/balena-cli/releases/download/v12.44.21/balena-cli-v12.44.21-windows-x64-installer.exe -s
Retrieving latest manifest for Balena.BalenaCLI
YamlDotNet.Core.YamlException: (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 32, Idx: 31): Exception during deserialization
 ---> System.InvalidCastException: Invalid cast from 'System.String' to 'Microsoft.WingetCreateCore.Models.ManifestTypeBase'.
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at System.String.System.IConvertible.ToType(Type type, IFormatProvider provider)
   at YamlDotNet.Serialization.Utilities.TypeConverter.ChangeType(Object value, Type destinationType, CultureInfo culture)
   at YamlDotNet.Serialization.Utilities.TypeConverter.ChangeType(Object value, Type destinationType)
   at YamlDotNet.Serialization.NodeDeserializers.ScalarNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   --- End of inner exception stack trace ---
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.Deserializer.Deserialize(IParser parser, Type type)
   at YamlDotNet.Serialization.Deserializer.Deserialize[T](IParser parser)
   at YamlDotNet.Serialization.Deserializer.Deserialize[T](TextReader input)
   at YamlDotNet.Serialization.Deserializer.Deserialize[T](String input)
   at Microsoft.WingetCreateCore.Serialization.DeserializeFromString[T](String value)
   at Microsoft.WingetCreateCore.Serialization.DeserializeManifestContents(IEnumerable`1 manifestContents, Manifests manifests)
   at Microsoft.WingetCreateCLI.Commands.UpdateCommand.DeserializeExistingManifestsAndUpdate(List`1 latestManifestContent)
```

# Expected behavior

Biip baap boop new manifest appears!

# Environment

>
```
Windows Manifest Creator v0.2.0.29
```

palenshus:
Fixes #90

Also modifying temporary GitHub clients used by the New and Update commands to use the passed-in GitHub token if present. the GitHub API has throttling which users may hit sometimes (I did when testing this fix), and we should have a way for users to get around that by specifying a token on the command line

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/91)

### DIFF
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -307,7 +307,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             VersionManifest versionManifest = manifests.VersionManifest;
             versionManifest.PackageIdentifier = PromptProperty(versionManifest, versionManifest.PackageIdentifier, nameof(versionManifest.PackageIdentifier));
 
-            GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+            GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
 
             string exactMatch = await client.FindPackageId(versionManifest.PackageIdentifier);
             if (!string.IsNullOrEmpty(exactMatch))
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -97,7 +97,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             {
                 Logger.DebugLocalized(nameof(Resources.RetrievingManifest_Message), this.Id);
 
-                GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+                GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
                 string exactId = await client.FindPackageId(this.Id);
 
                 if (!string.IsNullOrEmpty(exactId))
--- a/src/WingetCreateCore/Common/Serialization.cs
+++ b/src/WingetCreateCore/Common/Serialization.cs
@@ -146,29 +146,42 @@ namespace Microsoft.WingetCreateCore
         {
             foreach (string content in manifestContents)
             {
-                ManifestTypeBase baseType = Serialization.DeserializeFromString<ManifestTypeBase>(content);
+                string trimmedContent = RemoveBom(content);
+
+                ManifestTypeBase baseType = Serialization.DeserializeFromString<ManifestTypeBase>(trimmedContent);
 
                 if (baseType.ManifestType == "singleton")
                 {
-                    manifests.SingletonManifest = Serialization.DeserializeFromString<SingletonManifest>(content);
+                    manifests.SingletonManifest = Serialization.DeserializeFromString<SingletonManifest>(trimmedContent);
                 }
                 else if (baseType.ManifestType == "version")
                 {
-                    manifests.VersionManifest = Serialization.DeserializeFromString<VersionManifest>(content);
+                    manifests.VersionManifest = Serialization.DeserializeFromString<VersionManifest>(trimmedContent);
                 }
                 else if (baseType.ManifestType == "defaultLocale")
                 {
-                    manifests.DefaultLocaleManifest = Serialization.DeserializeFromString<DefaultLocaleManifest>(content);
+                    manifests.DefaultLocaleManifest = Serialization.DeserializeFromString<DefaultLocaleManifest>(trimmedContent);
                 }
                 else if (baseType.ManifestType == "locale")
                 {
-                    manifests.LocaleManifests.Add(Serialization.DeserializeFromString<LocaleManifest>(content));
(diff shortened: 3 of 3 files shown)


src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -97,7 +97,7 @@ public override async Task<bool> Execute()
             {
                 Logger.DebugLocalized(nameof(Resources.RetrievingManifest_Message), this.Id);
 
-                GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+                GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
```
ryfu-msft, line 100: If the user passes in a bad token, this will not only block the update functionality, but also throw a Octokit.AuthorizationException: Bad credentials exception. Are you okay with that behavior, we should at least catch the exception and show the user a proper error message IMO

src/WingetCreateCore/Common/Serialization.cs
```diff
@@ -146,31 +146,44 @@ public static void DeserializeManifestContents(IEnumerable<string> manifestConte
...
+        /// Removes Byte Order Marker (BOM) prefix if exists in string.
+        /// </summary>
+        /// <param name="value">String to fix.</param>
+        /// <returns>String without BOM prefix.</returns>
+        private static string RemoveBom(string value)
+        {
+            string bomMarkUtf8 = Encoding.UTF8.GetString(Encoding.UTF8.GetPreamble());
+            return value.StartsWith(bomMarkUtf8, StringComparison.OrdinalIgnoreCase) ? value.Remove(0, bomMarkUtf8.Length) : value;
+        }
+
```
ryfu-msft, line 186: Do you think it would be necessary to add a unit test for this? It should be pretty straightforward now that we can feed in local files to the updatecommand tests.

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -307,7 +307,7 @@ private async Task<bool> PromptPackageIdentifierAndCheckDuplicates(Manifests man
             VersionManifest versionManifest = manifests.VersionManifest;
             versionManifest.PackageIdentifier = PromptProperty(versionManifest, versionManifest.PackageIdentifier, nameof(versionManifest.PackageIdentifier));
 
-            GitHub client = new GitHub(null, this.WingetRepoOwner, this.WingetRepo);
+            GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
```
ryfu-msft, line 310: If the user passes in a bad token, this will fail before the user can create their manifest, is that what we want? I think for the new command this shouldn't be blocking, as we should just error out during the submission flow.

ryfu-msft (changes requested):

src/WingetCreateCore/Common/Serialization.cs
ryfu-msft: Discussed offline, not necessary since this basically is a trim.

ryfu-msft (approved):
:shipit:

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: Discussed offline. Current behavior in new, update, token, commands is to not catch this exception. This will just result in it getting thrown a little earlier if the user specifies a bad token

ryfu-msft: Resolved offline, if user enters a bad token, we should notify them immediately with a friendly message

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Resolved offline, if user enters a bad token, we should notify them immediately with a friendly message

PR #94, opened 2021-06-24
Add Telemetry Opt-Out and Settings Command

### LINKED ISSUE #84: Telemetry opt-out
At Microsoft, we care deeply about handling telemetry consent appropriately. As the winget-create tool does not have settings like the complementary Windows Package Manager, we are going to add a "first run" feature to provide an opt-out mechanism. This is in addition to the currently supported behavior associated with Windows telemetry. Even if the Windows settings permit telemetry, the winget-create tool will allow users to opt-out.

ryfu-msft:
This PR Resolves #84 and adds the following changes:

- Adds a `Settings` command that opens a settings json file similar to `winget settings`.
- Adds a settings json schema with a Telemetry.Disable field to allow for Telemetry opt-out. The schema allows for future optional configurations to be added.
- Adds a check when you first run the tool that asks the user to verify if they would like to enable telemetry.
- Adds documentation for the new `Settings` command
- The settings command works as follows:

1.  If no settings.json file exists, it will create one for you.
2.  If you already have a valid settings.json file, it would generate a settings.json.backup file to save your configurations in case settings.json ever breaks or becomes invalid.
3. If neither the settings.json or settings.json.backup file exist or are valid, it will use the default factory settings based on the settings schema.
4. Only running the `Settings` command will generate the proper settings files. For all other commands, a warning message will be shown, but it does change the settings files in any way.

Verification:
- Adds unit tests for the settings command to ensure that this behavior holds true.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/94)


src/WingetCreateCLI/Common.cs
```diff
@@ -2,27 +2,27 @@
 // Licensed under the MIT license.
 
 namespace Microsoft.WingetCreateCLI
```
palenshus, line 4: only whitespace changes in this file

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        /// <summary>
+        /// Setup method that removes the settings file and backup settings file prior to running tests.
+        /// </summary>
+        [SetUp]
+        public void SetUp()
+        {
+            if (File.Exists(UserSettings.SettingsJsonPath))
+            {
```
palenshus, line 35: You don't need these checks

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// Runs the settings command and modifies the file which should display a "Manifest cannot be empty." error.
+        /// </summary>
+        [Test]
+        public void VerifyEmptySettingsFile()
+        {
+            SettingsCommand command = new SettingsCommand();
+            command.Execute();
+            File.WriteAllText(UserSettings.SettingsJsonPath, string.Empty);
+            Assert.IsFalse(UserSettings.ParseJsonFile(UserSettings.SettingsJsonPath).IsValid, "Empty settings manifest should fail validation");
```
palenshus, line 80: >UserSettings.ParseJsonFile(UserSettings.SettingsJsonPath) [](http://example.com/codeflow?start=27&length=57)

Nit: Parse once and save to variable, then use in both asserts

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        /// <summary>
+        /// Gets the path to the backup settings json file.
+        /// </summary>
+        public static string SettingsBackupJsonPath => SettingsBackupPathLazy.Value;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to disable telemetry.
+        /// </summary>
+        public static bool TelemetryDisable
```
palenshus, line 42: >TelemetryDisable [](http://example.com/codeflow?start=27&length=16)

I believe this should be TelemetryDisabled

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                "Error message should be caught.");
+        }
+
+        /// <summary>
+        /// Compares the state of the telemetry.disabled field before and after to ensure the change is reflected accordingly.
+        /// </summary>
+        [Test]
+        public void VerifySavingTelemetrySettings()
+        {
+            bool isDisable = UserSettings.TelemetryDisable;
```
palenshus, line 93: >isDisable [](http://example.com/codeflow?start=17&length=9)

isDisabled

src/WingetCreateCLI/Telemetry/TelemetryEventListener.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Gets the static instance of the <see cref="TelemetryEventListener"/> class.
+        /// </summary>
+        public static TelemetryEventListener EventListener { get; } = new TelemetryEventListener();
+
+        /// <summary>
+        /// Checks if the Telemetry field of the settings file is and toggles telemetry based on that value.
```
palenshus, line 27: >Checks if the Telemetry field of the settings file is and toggles telemetry based on that value. [](http://example.com/codeflow?start=12&length=96)

reword

src/WingetCreateCLI/Telemetry/TelemetryEventListener.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Telemetry Event Listener class mainly used to disable telemetry events.
+    /// </summary>
+    public class TelemetryEventListener : EventListener
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TelemetryEventListener"/> class.
+        /// </summary>
+        public TelemetryEventListener()
```
palenshus, line 16: Shouldn't this be private?

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -556,4 +556,35 @@
...
     <value>We have detected that this package identifier already exists. If you intend to update an existing package, please use the update command.</value>
   </data>
+  <data name="EnableTelemetryFirstRun_Message" xml:space="preserve">
+    <value>Would you like to enable telemetry?</value>
+  </data>
+  <data name="SettingsCommand_HelpText" xml:space="preserve">
+    <value>Open settings</value>
+  </data>
+  <data name="ErrorParsingSettingsFile_Message" xml:space="preserve">
+    <value>Error parsing  file: {0}.</value>
```
palenshus, line 566: >  f [](http://example.com/codeflow?start=24&length=3)

two spaces

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -556,4 +556,35 @@
...
+    <comment>{0} - will be replaced with the path to the settings file.</comment>
+  </data>
+  <data name="GenerateNewSettingsFile_Message" xml:space="preserve">
+    <value>No settings file found, generating new settings file from loaded settings...</value>
```
palenshus, line 570: >generating new settings file from loaded settings... [](http://example.com/codeflow?start=35&length=52)

From loaded settings? Is that different than "default factory settings"?

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// <param name="path">Path to settings json file.</param>
+        /// <returns>A boolean IsValid that indicates if the file was parsed successfully and a list of error strings if the parsing fails.</returns>
+        public static (bool IsValid, List<string> Errors) ParseJsonFile(string path)
+        {
+            SettingsManifest settingsManifest;
+            List<string> errors = new List<string>();
+
+            using (StreamReader r = new StreamReader(path))
+            {
+                string json = r.ReadToEnd();
```
palenshus, line 70: Why not just File.ReadAllText?

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+                settingsManifest = JsonConvert.DeserializeObject<SettingsManifest>(
+                    json,
+                    new JsonSerializerSettings
+                    {
+                        Error = (sender, args) =>
+                        {
+                            errors.Add(args.ErrorContext.Error.Message);
+                            args.ErrorContext.Handled = true;
```
palenshus, line 78: nifty!

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+                        },
+                    });
+            }
+
+            if (settingsManifest == null)
+            {
+                errors.Add("Manifest cannot be empty.");
+            }
+
+            bool isValid = errors.Count == 0 && settingsManifest != null;
```
palenshus, line 88: > && settingsManifest != null [](http://example.com/codeflow?start=44&length=28)

redundant

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// Loads the correct settings file based on the following order.
+        /// 1. If the settings file exists and is valid, then overwrite the backup file with the current settings file.
+        /// 2. If the backup settings file exists and is valid, then recreate the settings file using the backup.
+        /// 3. Otherwise, use default settings based on the settings schema.
+        /// </summary>
+        private static void LoadSettings()
+        {
+            if (File.Exists(SettingsJsonPath) && ParseJsonFile(SettingsJsonPath).IsValid)
+            {
+                Settings = LoadJsonFile(SettingsJsonPath);
```
palenshus, line 143: If you want to avoid parsing the json file twice each time, you could add an out var to ParseJsonFile that returns the Settings object, and just assign that to Settings here, etc

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+            else if (File.Exists(SettingsBackupJsonPath) && ParseJsonFile(SettingsBackupJsonPath).IsValid)
+            {
+                Logger.WarnLocalized(nameof(Resources.UnexpectedErrorLoadSettings_Message));
+                Logger.WarnLocalized(nameof(Resources.LoadSettingsFromBackup_Message));
+                Settings = LoadJsonFile(SettingsBackupJsonPath);
+            }
+            else
+            {
+                // If either of these files exist, then this is not the first run and we can display warnings.
+                if (File.Exists(SettingsJsonPath) || File.Exists(SettingsBackupJsonPath))
```
palenshus, line 154: Comment doesn't seem to quite match scenario. I'd mention that we're here not just because the files exist, but because parsing them faile

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    Logger.WarnLocalized(nameof(Resources.UnexpectedErrorLoadSettings_Message));
+                    Logger.WarnLocalized(nameof(Resources.LoadSettingsFromDefault_Message));
+                }
+
+                Settings = new SettingsManifest { Telemetry = new Models.Settings.Telemetry() };
+            }
+        }
+
+        private static void SaveSettings()
```
palenshus, line 164: >SaveSettings [](http://example.com/codeflow?start=28&length=12)

This is almost identical to GenerateFileFromLoadedSettings no? I'd just get rid of the parameter and combine them

src/WingetCreateCLI/Commands/SettingsCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    (bool isBackupValid, List<string> backupFileErrors) = UserSettings.ParseJsonFile(UserSettings.SettingsBackupJsonPath);
+
+                    if (!isBackupValid)
+                    {
+                        this.DisplayParsingErrors(backupFileErrors, UserSettings.SettingsBackupJsonPath);
+                    }
+                }
+
+                return Task.FromResult(commandEvent.IsSuccessful = this.OpenJsonFile(UserSettings.SettingsJsonPath));
```
palenshus, line 66: >Task.FromResult( [](http://example.com/codeflow?start=23&length=16)

Why do you need this instead of just returning the bool?

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Common.cs
ryfu-msft: removed changes from this file.

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
ryfu-msft: Removed unnecessary checks

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
ryfu-msft: nice suggestion to helps save some code. fixed.

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: changed to TelemetryDisabled

src/WingetCreateTests/WingetCreateTests/UnitTests/SettingsCommandTests.cs
ryfu-msft: changed to isDisabled

src/WingetCreateCLI/Telemetry/TelemetryEventListener.cs
ryfu-msft: Reworded comment to more accurately reflect method

src/WingetCreateCLI/Telemetry/TelemetryEventListener.cs
ryfu-msft: yes you're right, changed to private.

src/WingetCreateCLI/Properties/Resources.resx
ryfu-msft: removed extra space

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: Changed to file.readalltext

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: thanks!

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: removed null check, now only checks that errors.Count is equal to 0

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: Added the out variable to the ParseJsonFile and removed LoadJsonFile method

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: Fixed comment to more accurately reflect that these warnings are due to parsing errors

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: Combined the methods by keeping SaveSettings and removing GeneratefileFromLoadedSettings

src/WingetCreateCLI/Commands/SettingsCommand.cs
ryfu-msft: This method is inherited from the BaseCommand and expects a Task<bool> to be returned. Since I don't have any async methods in this command, I have to resort to using Task.FromResult() in order to return a Task<bool>. 

Do you have a better alternative I can use?

palenshus: Ah I see. If you mark the method async you can just return the bool, but then I believe you'll get a warning saying that there's no async method calls, so I suppose this is probably preferable

palenshus (approved):

PR #100, opened 2021-06-29
Handle bad credential exception and block update submissions with no changes

### LINKED ISSUE #87: Update command shouldn't submit PR if there's no changes

When wingetcreate update is run with just a packageId, it'll fetch the package at the current URL, hash it, and submit an update PR. This makes sense if the URL is a vanity URL and the version is "latest", but not in other cases. For example, see this PR https://github.com/microsoft/winget-pkgs/pull/17720#issuecomment-862551629

# Steps to reproduce

wingetcreate update GitHub.cli -s

# Expected behavior

Aborts after determining that there's no substantial changes to be made

# Actual behavior

Downloads package at current URL, hashes it (resulting in the same hash), and submits a PR with no real changes

### LINKED ISSUE #93: Handle Bad Credentials Error with User-Friendly Message
Inputting a bad token shows an unfriendly error message. 

The correct behavior should should a user-friendly message that the token is invalid.

ryfu-msft:
The following pull request addresses several bugs listed below:

- Handles the bad credential exception by displaying a user-friendly error message to fix token. Fixes #93 
- Checks if hash is updated in update command and prevents submission if no change is detected. Fixes #87 
- Fixes an issue where the return code is nonzero when simply running `wingetcreate` with no arguments. 
- Fixes a bug in the telemetry for the New command where choosing to not submit would be classified as a command failure

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/100)


src/WingetCreateCLI/Program.cs
```diff
@@ -35,13 +37,15 @@ private static async Task<int> Main(string[] args)
...
+
+            var types = LoadVerbs();
+            var parserResult = myParser.ParseArguments(args, types);
+            BaseCommand command = parserResult.MapResult(c => c as BaseCommand, err => null);
+
             if (command == null)
             {
                 DisplayHelp(parserResult as NotParsed<object>);
-                return 1;
+                return (Environment.GetCommandLineArgs().Length > 1) ? 1 : 0;
```
palenshus, line 48: > (Environment.GetCommandLineArgs().Length > 1) [](http://example.com/codeflow?start=22&length=46)

args.Any()

src/WingetCreateCLI/Program.cs
```diff
@@ -61,7 +65,13 @@ private static async Task<int> Main(string[] args)
                 Logger.Error(ex.ToString());
                 return 1;
             }
-        }
+        }
+
+        private static Type[] LoadVerbs()
```
palenshus, line 70: "Load" to me implies more than what this is doing, I believe Get would be a better prefix.

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -150,18 +150,26 @@ public override async Task<bool> Execute()
 
                 string manifestDirectoryPath = SaveManifestDirToLocalPath(manifests, this.OutputDir);
 
-                if (!ValidateManifest(manifestDirectoryPath) ||
-                    !Prompt.Confirm(Resources.ConfirmGitHubSubmitManifest_Message) ||
-                    !await this.SetAndCheckGitHubToken())
-                {
+                bool isManifestValid;
```
palenshus, line 153: Why not just set the variable here instead of in the if clause?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -150,18 +150,26 @@ public override async Task<bool> Execute()
...
+
+                if (isManifestValid = ValidateManifest(manifestDirectoryPath) &&
+                    Prompt.Confirm(Resources.ConfirmGitHubSubmitManifest_Message))
+                {
+                    if (await this.SetAndCheckGitHubToken())
+                    {
+                        return commandEvent.IsSuccessful = await this.GitHubSubmitManifests(
```
palenshus, line 160: No need for this to be on three lines

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -95,9 +95,18 @@ public override async Task<bool> Execute()
 
             try
             {
+                GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
```
palenshus, line 98: Don't you want similar code in PromptPackageIdentifierAndCheckDuplicates?

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -192,27 +197,39 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
             }
 
-            string manifestDirectoryPath = SaveManifestDirToLocalPath(manifests, this.OutputDir);
+            string manifestDirectoryPath = SaveManifestDirToLocalPath(updatedManifests, this.OutputDir);
 
             if (ValidateManifest(manifestDirectoryPath))
             {
                 if (this.SubmitToGitHub)
                 {
+                    if (!this.CheckForUpdatedInstallerHash(originalManifests, updatedManifests))
```
palenshus, line 224: >CheckForUpdatedInstallerHash [](http://example.com/codeflow?start=30&length=28)

I'd change name to something like VerifyUpdated...

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -260,5 +277,41 @@ private static void UpdatePropertyForLocaleManifests(string propertyName, string
...
             }
         }
+
+        /// <summary>
+        /// Compares the hashes of the original and updated manifests and returns true if the new manifest has an updated SHA256 hash.
+        /// </summary>
+        /// <param name="oldManifests">The original manifest object model.</param>
+        /// <param name="newManifests">The updated manifest object model.</param>
+        /// <returns>A boolean value indicating whether the updated manifest has new changes compared to the original manifest.</returns>
+        private bool CheckForUpdatedInstallerHash(Manifests oldManifests, Manifests newManifests)
```
palenshus, line 287: I'd just pass the lists of installers directly since that's all this function needs

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -260,5 +277,41 @@ private static void UpdatePropertyForLocaleManifests(string propertyName, string
...
+                {
+                    continue;
+                }
+                else
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
palenshus, line 314: Learn to love LINQ :)

```
            var oldHashes = oldInstallerManifest.Installers.Select(i => i.InstallerSha256).Distinct();
            var newHashes = newInstallerManifest.Installers.Select(i => i.InstallerSha256).Distinct();
            return newHashes.Except(oldHashes).Any();

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 6 above)
Comment (palenshus) [inline src/WingetCreateCLI/Commands/UpdateCommand.cs:224]: >CheckForUpdatedInstallerHash [](http://example.com/codeflow?start=30&length=28)

I believe this will always succeed, since both have references to the same installers I think. Either way, you should add a UT for this

**Review by palenshus — DISMISSED** [review]
:clock1:

File: src/WingetCreateCLI/Program.cs (hunk 1 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Program.cs]: changed to suggestion

File: src/WingetCreateCLI/Program.cs (hunk 2 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Program.cs]: Changed to GetVerbs()

File: src/WingetCreateCLI/Commands/NewCommand.cs (hunk 3 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/NewCommand.cs]: changed to set variable before usage in if clause

File: src/WingetCreateCLI/Commands/NewCommand.cs (hunk 4 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/NewCommand.cs]: reformatted to single line

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 5 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs:98]: Let me know if you agree, but I think for PromptPackageIdentifierAndCheckDuplicates, I can just set that to null. There will never be a case when we want to be using the user's github token for the duplicate check. We would always want to check the public repo and therefore don't need a specific token.

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 6 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs]: Changed to VerifyUpdatedInstallerHash

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 7 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs]: I agree, changed to passing in installerManifest directly

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 8 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs]: truly magic :) Thanks for the suggestion.

File: src/WingetCreateCLI/Commands/UpdateCommand.cs (hunk 6 above)
Comment (ryfu-msft) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs]: Thanks for catching my mistake. I reverted the code back to using DeserializeExistingManifestsAndUpdate which has its own copy of the manifest object. I believe that this case should be covered by that UpdateAndVerifyUpdatedProperties() unit test. 
That unit test is comparing an original manifest with a new updated one and verifying the change. I believe that this should be enough to satisfy this case.

Comment (palenshus) [reply src/WingetCreateCLI/Commands/UpdateCommand.cs]: But that's a positive-only test, we should have a negative test too which ensures that if no hashes have changed after the update, then we fail

File: src/WingetCreateCLI/Commands/SubmitCommand.cs [hunk 9]
```diff
@@ -83,7 +83,7 @@ private async Task<bool> SubmitManifest()
             {
                 List<string> manifestContents = Directory.GetFiles(this.Path).Select(f => File.ReadAllText(f)).ToList();
 
-                Serialization.DeserializeManifestContents(manifestContents, manifests);
+                manifests = Serialization.DeserializeManifestContents(manifestContents);
```
palenshus, line 86: [nit] manifests variable isn't shared outside if-scopes, so line 75 could be moved into the first if block, and declared anew here, that way we don't construct it unnecessarily

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: We could definitely get away with this, but the reason I added the token to these calls was because GitHub kept throttling me after a few runs, so I had to provide a token to keep using the tool. If we get rid of it here, then the user will just have to wait until GitHub stops throttling them.

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/SubmitCommand.cs
ryfu-msft: Moved line 75 to first block and declare new Manifests object in 2nd block.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Good idea, I have added a negative test to make sure we check for this. This requires additional multifile test manifests just fyi

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Ahh i forgot about throttling. You're right, I added the token back and added a token check prior to prompting for packageId

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: Looks like this check won't apply/will fail if the source manifest is a singleton. Might make sense to factor out the singleton->multi manifest stuff so you can call it for the old and new manifests

ryfu-msft: I decided to add an additional check in the verifyinstallerhash method to look for the installer hash in the singleton manifest. That way we can still use our normal singleton test cases for the unit tests. lmk what you think

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -260,5 +276,23 @@ private static void UpdatePropertyForLocaleManifests(string propertyName, string
...
+
+        /// <summary>
+        /// Compares the hashes of the original and updated manifests and returns true if the new manifest has an updated SHA256 hash.
+        /// </summary>
+        /// <param name="oldManifest">The original manifest object model.</param>
+        /// <param name="newManifest">The updated installer manifest object model.</param>
+        /// <returns>A boolean value indicating whether the updated manifest has new changes compared to the original manifest.</returns>
+        private bool VerifyUpdatedInstallerHash(Manifests oldManifest, InstallerManifest newManifest)
+        {
+            IEnumerable<string> oldHashes;
```
palenshus, line 288: You can declare and init this var at once

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: Nice, that works well

src/WingetCreateTests/WingetCreateTests/TestConstants.cs
```diff
@@ -67,5 +67,20 @@ public static class TestConstants
...
+
+        /// <summary>
+        /// File name of the test multifile msi installer manifest.
+        /// </summary>
+        public const string TestMultifileMsiInstallerManifest = "Multifile.MsiTest.installer.yaml";
+
+        /// <summary>
+        /// File name of the test multifile msi default locale manifest.
+        /// </summary>
+        public const string TestMultifileMsiDefaultLocaleManifest = "Multifile.MsiTest.locale.en-US.yaml";
```
palenshus, line 84: Delete?

palenshus (approved):
:shipit:

PR #102, opened 2021-06-30
Adding support for multiple URLs in new and update command

### LINKED ISSUE #53: Multiple URL support

Add the possibility to add more than one installer in wingetcreate new and either wingetcreate update

~I think this is a low priority feature request, so keep calm and develop the rest 😆~
I didn't saw the source code, so I'm ignorant

palenshus:
Fixes #53 

Adds ability to specify multiple URLs on command-line to new and update commands. "new" is fairly straightforward, but "update" is much more complicated.

We require that the distinct URL count be the same between the old and new manifests, as well as the package count, and package types.

For example, you can update:

	x64 exe, x86 exe -> x64 exe, x86 exe
	msix (x64, x86) + x64 msi -> msix (x64, x86) + x64 msi

However, this will result in an error:

	x64 exe, x86 exe -> x64 exe
	msix (x64, x86) -> msix (x64, x86, arm64)

Any change to the installer set requires manual editing of the installer manifest.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/102)


palenshus:
E2E test will fail until WingetCreateE2E.MsixTest.yaml is updated in winget-pkgs-submission-test repo

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -1,299 +1,315 @@
...
+
+                Manifests manifests = new Manifests();
+
+                if (!this.InstallerUrls.Any())
+                {
+                    this.InstallerUrls = PromptProperty(
+                        new Installer(),
+                        this.InstallerUrls,
+                        nameof(Installer.InstallerUrl));
                 }
```
ryfu-msft, line 105: I believe this fails now because InstallerUrls is no longer just a string and is now a list of strings so there would be a casting error.

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -1,264 +1,285 @@
...
+            {
+                string packageFile = await DownloadPackageFile(url);
+                if (string.IsNullOrEmpty(packageFile))
+                {
+                    return null;
+                }
+
+                this.packageFiles.Add(packageFile);
+            }
+
```
ryfu-msft, line 186: I think this block of code is almost identical to the one in the NewCommand. You can refactor and put it in the basecommand class.

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -1,264 +1,285 @@
...
+            if (!PackageParser.UpdateInstallerNodes(installerManifest, this.InstallerUrls, this.packageFiles, out List<WingetCreateCore.Models.Installer.Installer> installersMissingMatches))
+            {
+                if (installersMissingMatches.Any())
+                {
+                    Logger.ErrorLocalized(nameof(Resources.MultipleInstallerUpdateDiscrepancy_Error));
+                    foreach (var installer in installersMissingMatches)
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.MissingPackageError_Message), installer.InstallerType, installer.Architecture);
```
ryfu-msft, line 194: >Resources.MissingPackageError_Message [](http://example.com/codeflow?start=53&length=37)

Make sure to add a comment in the resources file explaining what {0} and {1} will be replaced with.

src/WingetCreateCLI/Telemetry/Events/CommandExecutedEvent.cs
```diff
@@ -25,7 +26,7 @@ public class CommandExecutedEvent : EventBase
         /// <summary>
         /// Gets or sets the user provided installer URL.
         /// </summary>
-        public string InstallerUrl { get; set; }
+        public IEnumerable<string> InstallerUrls { get; set; }
```
ryfu-msft, line 30: Do you know if we can safely change the type of this? I believe I have this event registered with installerUrl as a string. Unless you can confirm that this is fine, I don't want to risk breaking the telemetry event

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -1,474 +1,540 @@
...
+        {
+            VersionManifest versionManifest = manifests.VersionManifest = new VersionManifest();
+
+            // TODO: Remove once default is set in schema
+            versionManifest.DefaultLocale = "en-US";
+
+            InstallerManifest installerManifest = manifests.InstallerManifest = new InstallerManifest();
+            DefaultLocaleManifest defaultLocaleManifest = manifests.DefaultLocaleManifest = new DefaultLocaleManifest();
+
+            foreach (var package in paths.Zip(urls, (path, url) => (path, url)))
```
ryfu-msft, line 82: >paths.Zip(urls, (path, url) => (path, url) [](http://example.com/codeflow?start=36&length=42)

super neat :)

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -1,474 +1,540 @@
...
+                {
+                    return false;
+                }
+            }
+
+            //// We only allow updating manifests with the same package count
+            //if (newInstallers.Count != existingInstallers.Count)
+            //{
+            //    return false;
+            //}
```
ryfu-msft, line 169: Can remove if not needed

ryfu-msft:
🕐

ryfu-msft (changes requested):
🕐

src/WingetCreateCore/Common/PackageParser.cs
palenshus: First time I've used it I think!

src/WingetCreateCLI/Commands/NewCommand.cs
palenshus: Nice catch! This was not fun to fix :) Unfortunately, sharprompt doesn't move the cursor down low enough after the user enters a big block of text, so the next output show up on top of the input text. I got around that by just console.clear'ing after this

src/WingetCreateCLI/Telemetry/Events/CommandExecutedEvent.cs
palenshus: Fair enough, joining to a single string

src/WingetCreateCLI/Telemetry/Events/CommandExecutedEvent.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace Microsoft.WingetCreateCLI.Telemetry.Events
 {
+    using System.Collections.Generic;
```
ryfu-msft, line 6: Can delete now that it's no longer needed

ryfu-msft (approved):
:shipit:

PR #106, opened 2021-07-07
Add prompts for optional string fields

ryfu-msft:
- Adds the basic prompting flow that prompts the user if they would like to add optional fields to their manifest. 
- This PR only prompts for basic string fields. 
- Also fixes the command line parser to handle -? without throwing an error.
- Moves the validation so that it occurs within the loop and prior to confirmation by the user.

The following optional fields are now customizable in the new command:
- PublisherUrl
- PublisherSupportUrl
- PrivacyUrl
- Author
- PackageUrl
- LicenseUrl
- Copyright
- CopyrightUrl
- Description
- Moniker
- Tags

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/106)


src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -332,7 +332,7 @@
     <value>The package meta-data locale</value>
   </data>
   <data name="PackageUrl_KeywordDescription" xml:space="preserve">
```
palenshus, line 334: Do we need examples in the resources for all of these as we do for the required properties? (@denelon / @KevinLaMS)

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -168,6 +164,23 @@ public override async Task<bool> Execute()
...
+        {
+            PromptRequiredProperties(manifests.VersionManifest);
+            PromptRequiredProperties(manifests.InstallerManifest, manifests.VersionManifest);
+            PromptRequiredProperties(manifests.DefaultLocaleManifest, manifests.VersionManifest);
+
+            Console.WriteLine();
+            if (Prompt.Confirm(Resources.AddOptionalFields_Message))
+            {
+                PromptOptionalProperties(manifests.DefaultLocaleManifest);
+                PromptOptionalProperties(manifests.InstallerManifest);
```
palenshus, line 178: I'm not sure if this is what we really want. This will prompt for optional properties on the InstallerManifest _root_, which includes properties which affect all installers (like MinimumOSVersion and InstallerLocale), but doesn't prompt for optional properties on the Installer nodes themselves.

Also, prompting only for string properties seems overly restrictive. For example, why can't the user set the Scope property here? That seems like one users would commonly want to do, since that's one of the more common optional properties.

Additionally, I don't think we should (ever?) prompt for some properties, like PackageFamilyName, which we can get from the MSIX, and ProductCode, which we can get from the MSI.

(@denelon / @KevinLaMS)

palenshus (changes requested):

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: So far this is the list of optional string fields that will be prompted. 
- Channel
- InstallerLocale
- MinimumOSVersion
- PackageFamilyName
- ProductCode
- PublisherUrl
- PublisherSupportUrl
- PrivacyUrl
- Author
- PackageUrl
- LicenseUrl
- Copyright
- CopyrightUrl
- Description
- Moniker

This is only the initial PR to allow users the ability to modify the most basic fields. Adding more complicated fields that take the form of enum, list<T>, object, will require more time to get those working properly. If we are trying to get the full functionality of optional fields working, then we will have to wait until next release. I do agree that some of the properties you pointed out should not be asked if we can parse them out.

ryfu-msft: An alternative would be to just prompt the values from the DefaultLocaleManifest and handle the optional fields from the installer manifest in a separate PR.

palenshus: I like that better as an initial implementation, that's much simpler to wrap my mind around. If we go that route, I'd also include Tags since that's another popular field. PromptProperty() already supports string[] properties, so you shouldn't have to do anything special for it, just make sure that the example mentions that they should be comma-delimited

denelon: Channel isn't implemented yet. I'd keep that one out.
I know Tags is an array so that may be worth holding off on, but that would be a high priority for moderators.

src/WingetCreateCLI/Properties/Resources.resx
denelon: I think it would be good to have examples, but I'd want to walk the UX to see how it looks. Some items are easier than others. The ones that are an enumeration might also be tricky like "user" & "machine" for scope.

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Thanks for the feedback! I've decided to only prompt optional fields from the defaultLocaleManifest. I've also spent some time working on adding Tags. I believe it works now and should work as expected.

src/WingetCreateCLI/Properties/Resources.resx
ryfu-msft: I think we should be careful adding too many examples, it could overload the screen and feel overwhelming since there will be quite a lot of text. I think this is a good question to figure out during our bug bash

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -215,6 +227,23 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
         }
 
+        private static void PromptOptionalProperties<T>(T manifest)
+        {
+            var properties = manifest.GetType().GetProperties().ToList();
+            var optionalProperties = properties.Where(p => p.GetCustomAttribute<RequiredAttribute>() == null).ToList();
+
+            foreach (var property in optionalProperties)
+            {
+                if (property.PropertyType == typeof(string) || property.Name == nameof(DefaultLocaleManifest.Tags))
```
palenshus, line 237: Do you still need this filter at all? If this is to avoid prompting for additionalProperties, you could change 233 to only include properties with the JsonProperty attribute too

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -300,12 +334,19 @@ private static T PromptProperty<T>(object model, T property, string memberName,
                     .MakeGenericMethod(property.GetType());
 
                 return (T)generic.Invoke(null, new object[] { message, property.GetType().GetEnumValues(), null, property, null });
-            }
-            else if (typeof(T) != typeof(string) && typeof(IEnumerable<string>).IsAssignableFrom(typeof(T)))
+            }
+            else if (typeof(IEnumerable<string>).IsAssignableFrom(typeof(T)) || typeof(IEnumerable<string>).IsAssignableFrom(typeFromModel))
```
palenshus, line 338: > || typeof(IEnumerable<string>).IsAssignableFrom(typeFromModel) [](http://example.com/codeflow?start=76&length=63)

What's a circumstance where this additional check would be required?

ryfu-msft: For optional fields that don't have existing values obtained from the package parser, we can't rely on typeof(T) to determine the type as they would be null and only show a type of System.Object.

The reason why we need both is because of the special case for installerURLs which isn't properly defined in the schema. It was supposed to only be defined as a string. All other List<string> fields are evaluated as lists and not individual strings. This is the rare case where validation actually considers the individual values of the list and not as a whole

palenshus: Discussed offline, there's times when typeof(T) would be Object. So flipping my question, when do we need the first clause here?

palenshus: Makes sense, thanks!

ryfu-msft: The first clause is needed for the InstallerURLs case. If we were to use reflection to determine the type of installerURL it would just be a string and not a List<string>. That is something we created to handle multiple installer urls.  Because we are passing in this.InstallerURLs directly to PromptProperty which is an IEnumerable, this is when the first clause would be true. 

There could be future cases where we start extrapolating out of the schema and start having lists of multiple items similar to the installerUrls.

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Added additional filter clause to filter out non-jsonproperty attributes

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -300,12 +334,19 @@ private static T PromptProperty<T>(object model, T property, string memberName,
...
-            else if (typeof(T) != typeof(string) && typeof(IEnumerable<string>).IsAssignableFrom(typeof(T)))
+            }
+            else if (typeof(IEnumerable<string>).IsAssignableFrom(typeof(T)) || typeof(IEnumerable<string>).IsAssignableFrom(typeFromModel))
             {
-                // If property is an IEnumerable<string>, we take in a comma-delimited string, and validate each split item, then return the split array
-                string promptResult = Prompt.Input<string>(message, null, new[] { FieldValidation.ValidateProperty(model, memberName, property) });
-                return (T)(object)promptResult?.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+                string combinedString = null;
+
+                if (property is List<string> propList && propList.Any())
```
palenshus, line 341: >List [](http://example.com/codeflow?start=32&length=4)

IEnumerable<string>

palenshus: Actually, isn't this first clause unnecessary, since we already know it's an IEnumerable<string> ?

src/WingetCreateCLI/FieldValidation.cs
```diff
@@ -31,40 +32,43 @@ public static class FieldValidation
...
-                if (defaultValue != null && defaultValue is not string && typeof(IEnumerable<string>).IsAssignableFrom(defaultValue.GetType()))
+                if (typeof(IEnumerable<string>).IsAssignableFrom(type) || (defaultValue != null && typeof(IEnumerable<string>).IsAssignableFrom(defaultValue.GetType())))
                 {
-                    // If prompt is an IEnumerable<string>, validate each item in the enumerable against the property
-
                     // If the user didn't provide a value, check if null is allowed for field
                     var items = property == null
-                        ? new string[] { null }
-                        : (property as string).Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+                        ? new List<string>()
```
palenshus, line 42: I may be missing why you changed this, but I needed the null in the array so that that value was checked against the field requirements on line 50, for InstallerUrls

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Changed to IEnumerable

src/WingetCreateCLI/FieldValidation.cs
ryfu-msft: I accidentally deleted the null, I also needed that. 

I changed the array to a list as I was having conflicts with the types when dealing with Tags which is a List<string>

palenshus (approved):
:shipit:

PR #108, opened 2021-07-13
Add Installer Url Matching for Update Command and Improve Error Messages

### LINKED ISSUE #105: Valid `wingetcreate update` fails with "New package found for Nullsoft X86 installer, but not found in existing manifest"

I am attempting to use wingetcreate to update [an existing manifest for SabakiHQ.Sabaki](https://github.com/microsoft/winget-pkgs/blob/master/manifests/s/SabakiHQ/Sabaki/0.51.1/SabakiHQ.Sabaki.installer.yaml). The update is to a new minor version that matches the same installer types as the existing one (they are produced by the same build pipeline).  However, the update fails after validating both provided installers with the following:

```powershell
PS C:\Users\apetresc\src\winget\winget-create> .\src\WingetCreateCLI\bin\x64\Debug\net5.0-windows10.0.17763.0\WingetCreateCLI.exe update SabakiHQ.Sabaki -v 0.52.0 -u "https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-x64-setup.exe" "https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-ia32-setup.exe"

Retrieving latest manifest for SabakiHQ.Sabaki
Downloading and parsing: https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-x64-setup.exe...

Downloading and parsing: https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-ia32-setup.exe...

Updating a manifest is only supported with the same number of installer URLs, and the same installer types and architectures
New package found for Nullsoft X86 installer, but not found in existing manifest
```
# Steps to reproduce

Simply the call above:
```powershell
PS C:\Users\apetresc\src\winget\winget-create> .\src\WingetCreateCLI\bin\x64\Debug\net5.0-windows10.0.17763.0\WingetCreateCLI.exe update SabakiHQ.Sabaki -v 0.52.0 -u "https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-x64-setup.exe" "https://github.com/SabakiHQ/Sabaki/releases/download/v0.52.0/sabaki-v0.52.0-win-ia32-setup.exe"
```

# Expected behavior

Since the provided URLs _do_ match the existing manifest's types and architecture (nullsoft x64 and x86 respectively), I expected `wingetcreate` to match them and generate an updated manifest.
# Actual behavior

It somehow fails to reconcile the installers even though they should match, and spits out that unhelpful error without providing any details of what it thinks the mismatch is.

# Environment

```none
PS C:\Users\apetresc\src\winget\winget-create> .\src\WingetCreateCLI\bin\x64\Debug\net5.0-windows10.0.17763.0\WingetCreateCLI.exe -?
Windows Manifest Creator v0.3.0.0
```
I'm using my own build of the latest `master` to work around #90 which hasn't been released yet.

### LINKED ISSUE #107: For multiple URL manifests, if two differ only by Scope (or presumably only by locale) then Update command will throw

PackageParser.cs:187, this checks for an existing Installer node to match to, using SingleOrDefault. If _no_ matches are found, it errors out appropriately. However, there may be cases where _multiple_ nodes are found, where multiple existing installers differ only by Scope or InstallerLocale. In those cases, SingleOrDefault will throw.

# Steps to reproduce

    wingetcreate update Microsoft.VisualStudioCode -u https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-ia32-1.59.0-insider.exe https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-ia32-1.59.0-insider.exe https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-x64-1.59.0-insider.exe https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-x64-1.59.0-insider.exe https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-arm64-1.59.0-insider.exe https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-arm64-1.59.0-insider.exe

# Expected behavior

Update command should fail with a helpful error indicating the reason

# Actual behavior

```
System.InvalidOperationException: Sequence contains more than one matching element
   at System.Linq.ThrowHelper.ThrowMoreThanOneMatchException()
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.WingetCreateCore.PackageParser.UpdateInstallerNodes(InstallerManifest installerManifest, IEnumerable`1 installerUrls, IEnumerable`1 paths, Boolean& installerMismatch, Installer& installerMissingMatch)
   at Microsoft.WingetCreateCLI.Commands.UpdateCommand.DeserializeExistingManifestsAndUpdate(List`1 latestManifestContent)
```

ryfu-msft:
This pull request addresses the following:

Fixes #105 
Fixes #107 

We have discovered that there are some installers that are incorrectly identified by their manifest. When looking at the installers, they may have an architecture or installer type that differs from what is specified in the manifest, leading to an update command failure due to mismatched installers. 

Since many installer URLs follow the pattern of including their architecture in their URL, we try to mitigate this issue by looking for any architectures stated in the installer URL. Winget-Create will attempt to match new installer URLs with their existing installers based on whether an architecture could be identified from the installer URL. If no architecture is identified, Winget-Create will resort to its normal flow and compare installers based on the architecture & installer type parsed from the package.

Changes:
- Fixes a bug where arm and arm64 were not being parsed from MSI installers
- Adds logic to check for a regex string match to find the architecture in the installer URL for matching updated installers.
- Improves error message for more clarity as to why an installer update might fail.

Testing:
- Adds a unit test to check whether the update command can determine the installer arch based on the installer URL using regex matching.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/108)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -263,17 +263,18 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// <param name="detectedArchs">List of DetectedArch objects that represent each installers detected architectures.</param>
         protected static void DisplayMismatchedArchitectures(List<PackageParser.DetectedArch> detectedArchs)
         {
-            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch).ToList();
+            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch);
+
             if (mismatchedArchInstallers.Any())
             {
                 Logger.WarnLocalized(nameof(Resources.DetectedArchMismatch_Message));
                 Console.WriteLine();
-                mismatchedArchInstallers.ForEach(i =>
+                foreach (var installer in mismatchedArchInstallers)
                 {
-                    Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), i.UrlArch, i.BinaryArch);
-                    Logger.Warn($"{i.Url}");
+                    Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), installer.UrlArch, installer.BinaryArch);
+                    Logger.Warn($"{installer.Url}");
                     Console.WriteLine();
-                });
+                }
             }
         }
 
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -260,13 +260,15 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// <summary>
         /// Displays the appropriate warning messages for installers with detected architecture mismatches.
         /// </summary>
-        protected static void DisplayMismatchedArchitectures()
+        /// <param name="detectedArchs">List of DetectedArch objects that represent each installers detected architectures.</param>
+        protected static void DisplayMismatchedArchitectures(List<PackageParser.DetectedArch> detectedArchs)
         {
-            if (PackageParser.ArchMismatches.Any())
+            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch).ToList();
+            if (mismatchedArchInstallers.Any())
             {
                 Logger.WarnLocalized(nameof(Resources.DetectedArchMismatch_Message));
                 Console.WriteLine();
-                PackageParser.ArchMismatches.ForEach(i =>
+                mismatchedArchInstallers.ForEach(i =>
                 {
                     Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), i.UrlArch, i.BinaryArch);
                     Logger.Warn($"{i.Url}");
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -268,7 +268,8 @@ namespace Microsoft.WingetCreateCLI.Commands
                 Console.WriteLine();
                 PackageParser.ArchMismatches.ForEach(i =>
                 {
-                    Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), i.Url, i.DetectedArch);
+                    Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), i.UrlArch, i.BinaryArch);
+                    Logger.Warn($"{i.Url}");
                     Console.WriteLine();
                 });
             }
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -232,6 +232,31 @@ namespace Microsoft.WingetCreateCLI.Commands
         }
 
         /// <summary>
+        /// Saves the manifests to a randomly generated directory in the %TEMP% folder and validates them, printing the results to console.
+        /// </summary>
+        /// <param name="manifests">Manifests object model.</param>
+        /// <returns>A boolean value indicating whether validation of the manifests was successful.</returns>
+        protected static bool ValidateManifestsInTempDir(Manifests manifests)
+        {
+            string packageId = manifests.VersionManifest.PackageIdentifier;
+            string defaultPackageLocale = manifests.DefaultLocaleManifest.PackageLocale;
+            string versionManifestFileName = $"{packageId}.yaml";
+            string installerManifestFileName = $"{packageId}.installer.yaml";
+            string defaultLocaleManifestFileName = $"{packageId}.locale.{defaultPackageLocale}.yaml";
+
+            string randomDirPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            Directory.CreateDirectory(randomDirPath);
+
+            File.WriteAllText(Path.Combine(randomDirPath, versionManifestFileName), manifests.VersionManifest.ToYaml());
+            File.WriteAllText(Path.Combine(randomDirPath, installerManifestFileName), manifests.InstallerManifest.ToYaml());
+            File.WriteAllText(Path.Combine(randomDirPath, defaultLocaleManifestFileName), manifests.DefaultLocaleManifest.ToYaml());
+
+            bool result = ValidateManifest(randomDirPath);
+            Directory.Delete(randomDirPath, true);
+            return result;
+        }
+
+        /// <summary>
         /// Launches the OAuth flow to allow the user to login to their GitHub account and grant permission to the app.
         /// </summary>
         /// <returns>Access token string.</returns>
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -6,6 +6,7 @@ namespace Microsoft.WingetCreateCLI.Commands
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
     using System.Net.Http;
     using System.Threading.Tasks;
     using CommandLine;
@@ -254,6 +255,23 @@ namespace Microsoft.WingetCreateCLI.Commands
             bool result = ValidateManifest(randomDirPath);
             Directory.Delete(randomDirPath, true);
             return result;
+        }
+
+        /// <summary>
+        /// Displays the appropriate warning messages for installers with detected architecture mismatches.
+        /// </summary>
+        protected static void DisplayMismatchedArchitectures()
+        {
+            if (PackageParser.ArchMismatches.Any())
+            {
+                Logger.WarnLocalized(nameof(Resources.DetectedArchMismatch_Message));
+                Console.WriteLine();
+                PackageParser.ArchMismatches.ForEach(i =>
+                {
+                    Logger.WarnLocalized(nameof(Resources.InstallerBinaryMismatch_Message), i.Url, i.DetectedArch);
+                    Console.WriteLine();
+                });
+            }
         }
 
         /// <summary>
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -263,7 +263,7 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// <param name="detectedArchs">List of DetectedArch objects that represent each installers detected architectures.</param>
         protected static void DisplayMismatchedArchitectures(List<PackageParser.DetectedArch> detectedArchs)
         {
-            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch);
+            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch.HasValue && i.UrlArch != i.BinaryArch);
 
             if (mismatchedArchInstallers.Any())
             {
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -199,35 +199,30 @@ namespace Microsoft.WingetCreateCLI.Commands
                 this.InstallerUrls,
                 packageFiles,
                 out bool installerMismatch,
-                out bool multipleMatchesFound,
-                out List<WingetCreateCore.Models.Installer.Installer> installersMissingMatch))
+                out List<WingetCreateCore.Models.Installer.Installer> unmatchedInstallers,
+                out List<WingetCreateCore.Models.Installer.Installer> multipleMatchedInstallers))
             {
                 if (installerMismatch)
                 {
                     Logger.ErrorLocalized(nameof(Resources.MultipleInstallerUpdateDiscrepancy_Error));
                     Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
 
-                    if (installersMissingMatch.Any())
+                    if (unmatchedInstallers.Any())
                     {
                         Logger.ErrorLocalized(nameof(Resources.InstallerMatchFailedError_Message));
+                        unmatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
                     }
-                }
-                else if (multipleMatchesFound)
-                {
-                    Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
+
+                    if (multipleMatchedInstallers.Any())
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
+                        multipleMatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
+                    }
                 }
                 else
                 {
                     Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
                 }
-
-                installersMissingMatch.ForEach(i =>
-                {
-                    if (installerMismatch || multipleMatchesFound)
-                    {
-                        Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl);
-                    }
-                });
 
                 return null;
             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -3,6 +3,7 @@
 
 namespace Microsoft.WingetCreateCLI.Commands
 {
+    using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
@@ -204,25 +205,32 @@ namespace Microsoft.WingetCreateCLI.Commands
             {
                 if (installerMismatch)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.MultipleInstallerUpdateDiscrepancy_Error));
                     Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
 
                     if (unmatchedInstallers.Any())
                     {
                         Logger.ErrorLocalized(nameof(Resources.InstallerMatchFailedError_Message));
-                        unmatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
+                        Console.WriteLine();
+                        unmatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
                     }
 
                     if (multipleMatchedInstallers.Any())
                     {
                         Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
-                        multipleMatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
+                        Console.WriteLine();
+                        multipleMatchedInstallers.ForEach(i =>
+                        {
+                            Logger.ErrorLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl);
+                            Console.WriteLine();
+                        });
                     }
                 }
                 else
                 {
                     Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
                 }
+
+                DisplayMismatchedArchitectures();
 
                 return null;
             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -201,7 +201,8 @@ namespace Microsoft.WingetCreateCLI.Commands
                 packageFiles,
                 out bool installerMismatch,
                 out List<WingetCreateCore.Models.Installer.Installer> unmatchedInstallers,
-                out List<WingetCreateCore.Models.Installer.Installer> multipleMatchedInstallers))
+                out List<WingetCreateCore.Models.Installer.Installer> multipleMatchedInstallers,
+                out List<PackageParser.DetectedArch> detectedArchs))
             {
                 if (installerMismatch)
                 {
@@ -230,7 +231,7 @@ namespace Microsoft.WingetCreateCLI.Commands
                     Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
                 }
 
-                DisplayMismatchedArchitectures();
+                DisplayMismatchedArchitectures(detectedArchs);
 
                 return null;
             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -202,7 +202,7 @@ namespace Microsoft.WingetCreateCLI.Commands
                 out bool installerMismatch,
                 out List<WingetCreateCore.Models.Installer.Installer> unmatchedInstallers,
                 out List<WingetCreateCore.Models.Installer.Installer> multipleMatchedInstallers,
-                out List<PackageParser.DetectedArch> detectedArchs))
+                out List<PackageParser.DetectedArch> detectedArchOfInstallers))
             {
                 if (installerMismatch)
                 {
@@ -231,7 +231,7 @@ namespace Microsoft.WingetCreateCLI.Commands
                     Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
                 }
 
-                DisplayMismatchedArchitectures(detectedArchs);
+                DisplayMismatchedArchitectures(detectedArchOfInstallers);
 
                 return null;
             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -194,20 +194,40 @@ namespace Microsoft.WingetCreateCLI.Commands
                 return null;
             }
 
-            if (!PackageParser.UpdateInstallerNodes(installerManifest, this.InstallerUrls, packageFiles, out bool installerMismatch, out WingetCreateCore.Models.Installer.Installer installersMissingMatch))
+            if (!PackageParser.UpdateInstallerNodesAsync(
+                installerManifest,
+                this.InstallerUrls,
+                packageFiles,
+                out bool installerMismatch,
+                out bool multipleMatchesFound,
+                out List<WingetCreateCore.Models.Installer.Installer> installersMissingMatch))
             {
                 if (installerMismatch)
                 {
                     Logger.ErrorLocalized(nameof(Resources.MultipleInstallerUpdateDiscrepancy_Error));
-                    if (installersMissingMatch != null)
+                    Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
+
+                    if (installersMissingMatch.Any())
                     {
-                        Logger.ErrorLocalized(nameof(Resources.MissingPackageError_Message), installersMissingMatch.InstallerType, installersMissingMatch.Architecture);
+                        Logger.ErrorLocalized(nameof(Resources.InstallerMatchFailedError_Message));
                     }
+                }
+                else if (multipleMatchesFound)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
                 }
                 else
                 {
                     Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
                 }
+
+                installersMissingMatch.ForEach(i =>
+                {
+                    if (installerMismatch || multipleMatchesFound)
+                    {
+                        Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl);
+                    }
+                });
 
                 return null;
             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -268,7 +268,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             {
                 if (this.SubmitToGitHub)
                 {
-                    if (!this.VerifyUpdatedInstallerHash(originalManifests, updatedManifests.InstallerManifest))
+                    if (!VerifyUpdatedInstallerHash(originalManifests, updatedManifests.InstallerManifest))
                     {
                         Logger.ErrorLocalized(nameof(Resources.NoChangeDetectedInUpdatedManifest_Message));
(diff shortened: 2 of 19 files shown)


src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,17 +191,33 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
             // Update previous installers with parsed data from downloaded packages
             foreach (var newInstaller in newInstallers)
             {
-                // Find a match for the current installer in the list of installers remaining to be updated.
-                var matchingExistingInstaller = existingInstallers.SingleOrDefault(i =>
-                    (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
-                    i.Architecture == newInstaller.Architecture);
+                var archGuess = GetArchFromUrl(newInstaller.InstallerUrl);
+                dynamic matchingExistingInstaller = null;
```
palenshus, line 195: >dynamic [](http://example.com/codeflow?start=16&length=7)

Make Installer

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -209,7 +233,24 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
-            return true;
+            return !installerMissingMatch.Any();
+        }
+
+        /// <summary>
+        /// Performs a regex match to determine the installer architecture based on the url string.
+        /// </summary>
+        /// <param name="url">Installer url string.</param>
+        /// <returns>Installer architecture enum.</returns>
+        private static InstallerArchitecture? GetArchFromUrl(string url)
```
palenshus, line 244: >GetArchFromUrl [](http://example.com/codeflow?start=46&length=14)

Use new logic from linqpad script

palenshus:
Check Firefox to see how we handle apps that differ only by locale

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: fixed

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Changed to newest logic from linqpad script

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -186,7 +186,26 @@ public async Task UpdateFailsWithUnmatchedPackages()
...
+            TestUtils.InitializeMockDownloads(TestConstants.TestMsiInstaller);
+            (UpdateCommand command, var initialManifestContent) = GetUpdateCommandAndManifestData("TestPublisher.MatchWithInstallerUrl", null, this.tempPath, null);
+            var initialManifests = Serialization.DeserializeManifestContents(initialManifestContent);
+            var initialInstaller = initialManifests.SingletonManifest.Installers.First();
+
+            var updatedManifests = await command.DeserializeExistingManifestsAndUpdate(initialManifestContent);
+            Assert.IsNotNull(updatedManifests, "Command should have succeeded");
+            Assert.AreNotEqual(initialInstaller.InstallerSha256 != updatedManifests.InstallerManifest.Installers.FirstOrDefault().InstallerSha256, "Installer hash should be updated.");
+            TestUtils.TestInstallerBaseUrl = TestUtils.DefaultTestInstallerBaseUrl;
         }
```
palenshus, line 209: Safer to do this in the Teardown method, since if this test fails out early, it'll corrupt the rest of the tests run after it

palenshus:
> Check Firefox to see how we handle apps that differ only by locale

@ryfu-msft, how'd this go?

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -194,14 +194,29 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
+                installerManifest,
+                this.InstallerUrls,
+                packageFiles,
+                out bool installerMismatch,
+                out List<WingetCreateCore.Models.Installer.Installer> unmatchedInstallers,
+                out List<WingetCreateCore.Models.Installer.Installer> multipleMatchedInstallers))
             {
                 if (installerMismatch)
                 {
                     Logger.ErrorLocalized(nameof(Resources.MultipleInstallerUpdateDiscrepancy_Error));
```
palenshus, line 207: We shouldn't log this error here, as we've already verified that the installer URL count matches (my bad)

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -194,14 +194,29 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.InstallerMatchFailedError_Message));
+                        unmatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
+                    }
+
+                    if (multipleMatchedInstallers.Any())
                     {
-                        Logger.ErrorLocalized(nameof(Resources.MissingPackageError_Message), installersMissingMatch.InstallerType, installersMissingMatch.Architecture);
+                        Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
+                        multipleMatchedInstallers.ForEach(i => Logger.WarnLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl));
```
palenshus, line 219: I tried VSCode and found this message to be a bit misleading.

```
Each new installer URL must have a match to an existing installer node based on installer type and architecture.
Multiple existing installer nodes matched the following installer(s):
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-ia32-1.59.0-insider.exe
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-ia32-1.59.0-insider.exe
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-x64-1.59.0-insider.exe
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-x64-1.59.0-insider.exe
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeUserSetup-arm64-1.59.0-insider.exe
X86 Inno installer detected from the url: https://az764295.vo.msecnd.net/insider/87d692b7bf1eb5a663b3e62620a6e7e436e1838c/VSCodeSetup-arm64-1.59.0-insider.exe
```

This makes it seem like maybe the problem is that we detected x86 for all the installers, when in fact with our new URL parsing logic, we actually detect the correct architectures, so showing this makes it seem like a different issue. Perhaps we need to update the Installer node with the arch detected from the URL prior to bailing out of the UpdateInstallerNodesAsync method?

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,17 +191,46 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
             // Update previous installers with parsed data from downloaded packages
             foreach (var newInstaller in newInstallers)
             {
-                // Find a match for the current installer in the list of installers remaining to be updated.
-                var matchingExistingInstaller = existingInstallers.SingleOrDefault(i =>
-                    (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
+                var archGuess = GetArchFromUrl(newInstaller.InstallerUrl);
+
```
palenshus, line 195: Log a warning here if our archGuess != newInstaller.Architecture, this'll let the user know we're overriding what was detected in the binary, and possibly spur them to fix the binary arch too...

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,17 +191,46 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
             // Update previous installers with parsed data from downloaded packages
             foreach (var newInstaller in newInstallers)
             {
-                // Find a match for the current installer in the list of installers remaining to be updated.
-                var matchingExistingInstaller = existingInstallers.SingleOrDefault(i =>
-                    (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
+                var archGuess = GetArchFromUrl(newInstaller.InstallerUrl);
```
palenshus, line 194: Now that we have this cool logic, I wonder if we should use it in the new command too? For example, this currently detects the wrong arch type:
```
new https://download.videolan.org/vlc/3.0.15/win64/vlc-3.0.15-win64.exe https://download.videolan.org/vlc/3.0.15/win32/vlc-3.0.15-win32.exe
```
What about adding the override after PackageParser:337 too? (And log a warning if we detect a mismatch)

palenshus (dismissed):

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: moved to teardown method

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: > > Check Firefox to see how we handle apps that differ only by locale
> 
> @ryfu-msft, how'd this go?

As expected, installer nodes that differed only by locale had a similar behavior to installer nodes that only differed by scope. We will need to investigate the heuristics to see if we can catch these cases.

ryfu-msft:
> > Check Firefox to see how we handle apps that differ only by locale
> 
> @ryfu-msft, how'd this go?

As expected, installer nodes that differed only by locale had a similar behavior to installer nodes that only differed by scope. We will need to investigate the heuristics to see if we can catch these cases.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Removed logging here

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: I agree that this can cause some confusions. I added some extra logic and messages so that it is clear to the user what the differences are between architecture detections.

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Added the cool logic to the new command. New command now shows the warning messages if a arch mismatch is detected prior to prompting for required fields.

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Added a log message that tells the user the detected arch in the binary differs from what what detected in the URL. Let me know what you think and if it needs more clarification.

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,17 +197,52 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
+
+                if (archGuess != null && archGuess != newInstaller.Architecture)
+                {
+                    ArchMismatches.Add((newInstaller.InstallerUrl, (InstallerArchitecture)archGuess, newInstaller.Architecture));
```
palenshus, line 204: >(InstallerArchitecture) [](http://example.com/codeflow?start=67&length=23)

Use archGuess.Value instead

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -209,7 +258,40 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
+            installerMismatch = unmatchedInstallers.Any() || multipleMatchedInstallers.Any();
+            return !installerMismatch;
+        }
+
+        /// <summary>
+        /// Performs a regex match to determine the installer architecture based on the url string.
+        /// </summary>
+        /// <param name="url">Installer url string.</param>
+        /// <returns>Installer architecture enum.</returns>
+        public static InstallerArchitecture? GetArchFromUrl(string url)
```
palenshus, line 270: This doesn't need to be public anymore

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -255,13 +337,25 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
+            string url,
+            List<Installer> installers,
+            Manifests manifests)
         {
             var installer = new Installer();
             installer.InstallerUrl = url;
             installer.InstallerSha256 = GetFileHash(path);
             installer.Architecture = GetMachineType(path)?.ToString().ToEnumOrDefault<InstallerArchitecture>() ?? InstallerArchitecture.Neutral;
 
+            var archGuess = GetArchFromUrl(installer.InstallerUrl);
```
palenshus, line 351: At this point, we've only tried to parse the arch from the exe, not an MSI. This needs to be after line 361

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,17 +197,52 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
-                // Find a match for the current installer in the list of installers remaining to be updated.
-                var matchingExistingInstaller = existingInstallers.SingleOrDefault(i =>
-                    (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
+                var archGuess = GetArchFromUrl(newInstaller.InstallerUrl);
+
+                if (archGuess != null && archGuess != newInstaller.Architecture)
+                {
+                    ArchMismatches.Add((newInstaller.InstallerUrl, (InstallerArchitecture)archGuess, newInstaller.Architecture));
+                    newInstaller.Architecture = (InstallerArchitecture)archGuess;
+                }
```
palenshus, line 206: Doesn't this mess up the logging for unmatched/multipleMatchedInstallers? I believe you need to return a tuple for those too...

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -256,6 +257,24 @@ protected static bool ValidateManifestsInTempDir(Manifests manifests)
             return result;
         }
 
+        /// <summary>
+        /// Displays the appropriate warning messages for installers with detected architecture mismatches.
+        /// </summary>
+        protected static void DisplayMismatchedArchitectures()
+        {
+            if (PackageParser.ArchMismatches.Any())
+            {
```
palenshus, line 266: I don't like the idea of making PackageParser stateful. Is there any reason you can't just have DisplayMismatchedArchitectures take the mismatches as a parameter?

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -611,4 +609,24 @@
...
+    <value>Multiple existing installer nodes matched the following installer(s):</value>
+  </data>
+  <data name="NewInstallerUrlMustMatchExisting_Message" xml:space="preserve">
+    <value>Each new installer URL must have a match to an existing installer node based on installer type and architecture.</value>
+  </data>
+  <data name="DetectedArchMismatch_Message" xml:space="preserve">
+    <value>The detected architecture might be different than what is specified in the installer URL for the following installers:</value>
+  </data>
+  <data name="InstallerBinaryMismatch_Message" xml:space="preserve">
+    <value>Selected architecture detected from URL: {0} | Architecture detected from binary: {1}</value>
```
palenshus, line 628: I think I'd change this to something like:

Using architecture detected from URL ({0}), overriding architecture detected from binary ({1})

What do you think?

palenshus (changes requested):
:clock1:

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: fixed

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: changed to private

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: I missed this. moved logic to after line 361

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Yes you're right, I created a detectedArch object to keep track of the architectures detected from the url and binary. I believe the logging should be correct now that the the different detected architectures are stored separately

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Changed displaymismatchedarchitectures to take in a parameter. Created an out variable to pass along the detected architectures.

src/WingetCreateCLI/Properties/Resources.resx
ryfu-msft: I like that better, more clear about the fact that we overrode the architecture. Change to suggested.

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -548,5 +644,39 @@ private static string RemoveInvalidCharsFromString(string value)
         {
             return Regex.Replace(value, InvalidCharacters, string.Empty);
         }
+
+        /// <summary>
+        /// DetectedArch object with information about an installer's architectures detected from the url and binary.
+        /// </summary>
+        public class DetectedArch
```
palenshus, line 657: >DetectedArch [](http://example.com/codeflow?start=21&length=12)

You can use a record type for this (see PublisherApp.cs)

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -255,16 +333,37 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
             installer.Architecture = GetMachineType(path)?.ToString().ToEnumOrDefault<InstallerArchitecture>() ?? InstallerArchitecture.Neutral;
 
-            return ParseExeInstallerType(path, installer, installers) ||
+            bool parseResult = ParseExeInstallerType(path, installer, installers) ||
                 ParseMsix(path, installer, manifests, installers) ||
                 ParseMsi(path, installer, manifests, installers);
+
+            var archGuess = GetArchFromUrl(installer.InstallerUrl);
+
+            if (!archGuess.HasValue)
```
palenshus, line 354: You can shorten this to var archGuess = GetArchFromUrl(..) ?? installer.InstallerUrl;

palenshus: That said, why set this at all? It's not a guess anymore if you overwrite it?

palenshus: I think this could all be re-written as:

            var archGuess = GetArchFromUrl(installer.InstallerUrl);
            detectedArchOfInstallers.Add(new DetectedArch(installer.InstallerUrl, archGuess.Value, installer.Architecture));

            if (archGuess.HasValue)
            {
                installer.Architecture = archGuess.Value;
            }

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -255,16 +333,37 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
             }
         }
 
-        private static bool ParsePackageAndGenerateInstallerNodes(string path, string url, List<Installer> installers, Manifests manifests)
+        private static bool ParsePackageAndGenerateInstallerNodes(
+            string path,
+            string url,
+            List<Installer> installers,
+            Manifests manifests,
+            ref List<DetectedArch> detectedArchOfInstallers)
```
palenshus, line 341: >detectedArchOfInstallers [](http://example.com/codeflow?start=35&length=24)

Use same variable name everywhere

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Because GetArchFromUrl returns a nullable type, archGuess.Value will throw an invalidoperationexception if there is no value.

src/WingetCreateCore/Common/PackageParser.cs
palenshus: Only for exe/msi. MSIX we shouldn't override

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -256,6 +257,26 @@ protected static bool ValidateManifestsInTempDir(Manifests manifests)
...
+        /// </summary>
+        /// <param name="detectedArchs">List of DetectedArch objects that represent each installers detected architectures.</param>
+        protected static void DisplayMismatchedArchitectures(List<PackageParser.DetectedArch> detectedArchs)
+        {
+            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch).ToList();
+            if (mismatchedArchInstallers.Any())
+            {
+                Logger.WarnLocalized(nameof(Resources.DetectedArchMismatch_Message));
+                Console.WriteLine();
+                mismatchedArchInstallers.ForEach(i =>
```
palenshus, line 271: nit: I'd use foreach syntax here, it's more common. You also don't need the .ToList() if you do that

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -256,6 +257,26 @@ protected static bool ValidateManifestsInTempDir(Manifests manifests)
             return result;
         }
 
+        /// <summary>
+        /// Displays the appropriate warning messages for installers with detected architecture mismatches.
+        /// </summary>
+        /// <param name="detectedArchs">List of DetectedArch objects that represent each installers detected architectures.</param>
+        protected static void DisplayMismatchedArchitectures(List<PackageParser.DetectedArch> detectedArchs)
+        {
+            var mismatchedArchInstallers = detectedArchs.Where(i => i.UrlArch != i.BinaryArch).ToList();
```
palenshus, line 266: I'd say this should be where i.UrlArch.HasValue && i.UrlArch != i.BinaryArch

src/WingetCreateCore/Common/PackageParser.cs
palenshus: Ah, so just archGuess then, and DetectedArch.UrlArch should be nullable

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: made changes so that architecture override only occurs for exe/msi and not MSIX

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Changed to record type

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Changed to same variable name

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: changed to foreach syntax

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Changed to suggested with DetectedArch.UrlArch as nullable

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Changed to suggested

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -255,16 +340,47 @@ public static bool UpdateInstallerNodes(InstallerManifest installerManifest, IEn
...
+
+            if (parseMsixResult)
+            {
+                // Skip architecture override for msix installers as there can be more than one installer in a bundle
+                var nonMsixInstallerUrls = detectedArchOfInstallers.Select(i => i.Url).ToList();
+                var msixInstallers = installers.Where(i => !nonMsixInstallerUrls.Contains(i.InstallerUrl));
+                foreach (var msixInstaller in msixInstallers)
+                {
+                    var msixArch = msixInstaller.Architecture;
+                    detectedArchOfInstallers.Add(new DetectedArch(msixInstaller.InstallerUrl, msixArch, msixArch));
```
palenshus, line 369: I don't think you need to add/log these at all? Since we're not overriding these, we don't need to worry about warning the user about them, right?

ryfu-msft: I know this is slightly confusing but I need to add these to detectedArchOfInstallers or else matching will fail on PackageParser Line 205 - 215. This is because ParseMsix adds new installers from the bundle, which is not picked up by detectedArchOfInstallers so when it tries to match by installerURL and Arch, it fails. since those installers aren't present. 

These won't get shown in the logs because urlArch == binaryArch so logging isn't affected.

palenshus: detectedArchOfInstallers.AddRange(
    installers
    .Where(i => i.InstallerUrl == url)
    .Select(i => new DetectedArch(i.InstallerUrl, i.Architecture, i.Architecture))
)

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -185,8 +186,26 @@ public async Task UpdateFailsWithUnmatchedPackages()
...
+        /// <summary>
+        /// Since some installers are incorrectly labeled on the manifest, resort to using the installer URL to find matches.
+        /// This unit test uses a msi installer that is not an arm64 installer, but because the installer URL includes "arm64", it should find a match.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task UpdateBasedOnInstallerUrlMatch()
+        {
+            TestUtils.TestInstallerBaseUrl = "https://fakedomain.com/arm64/";
+            TestUtils.InitializeMockDownloads(TestConstants.TestMsiInstaller);
```
palenshus, line 201: You can pass in multiple URLs by changing InitializeMockdownloads

src/WingetCreateCore/Common/PackageParser.cs
palenshus: Test

ryfu-msft: Changed to suggestion

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: Reverted to what we had before. I decided to go with your suggestion about using Path.GetFileName() to separate out the architecture from the filename.

src/WingetCreateTests/WingetCreateTests/UnitTests/PackageParserTests.cs
```diff
@@ -48,7 +49,7 @@ public void ParseExeInstallerFile()
 
             Manifests manifests = new Manifests();
 
-            Assert.IsTrue(PackageParser.ParsePackages(new[] { testExeInstallerPath }, new[] { TestConstants.TestExeInstaller }, manifests));
+            Assert.IsTrue(PackageParser.ParsePackages(new[] { testExeInstallerPath }, new[] { TestConstants.TestExeInstaller }, manifests, out List<PackageParser.DetectedArch> detectedArchOfInstallers));
```
palenshus, line 52: Since these out vars are never used, use a discard instead:

, out _)

palenshus:
>         private bool VerifyUpdatedInstallerHash(Manifests oldManifest, InstallerManifest newManifest)

This can be static

---
Refers to: src/WingetCreateCLI/Commands/UpdateCommand.cs:334 in 1000920. [](commit_id = 10009203810225fff473a4ca146fa95b8ef021d7, deletion_comment = False)

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -185,8 +185,53 @@ public async Task UpdateFailsWithUnmatchedPackages()
...
+                "win32/" + TestConstants.TestMsiInstaller);
+            (UpdateCommand command, var initialManifestContent) = GetUpdateCommandAndManifestData("TestPublisher.MatchWithInstallerUrl", null, this.tempPath, null);
+            var initialManifests = Serialization.DeserializeManifestContents(initialManifestContent);
+            var initialInstaller = initialManifests.SingletonManifest.Installers.First();
+
+            var updatedManifests = await command.DeserializeExistingManifestsAndUpdate(initialManifestContent);
+            Assert.IsNotNull(updatedManifests, "Command should have succeeded");
+            foreach (var updatedInstaller in updatedManifests.InstallerManifest.Installers)
+            {
+                Assert.AreNotEqual(initialInstaller.InstallerSha256, updatedInstaller.InstallerSha256, "InstallerSha256 should be updated");
```
palenshus, line 212: Instead of just checking !=, let's check if each one matches what is expected, that way we can make sure the URL parsing logic works. You can do this by making the archs an array:

var archs = new[] { "arm64", "arm"... };
TestUtils.InitializeMockDownloads(archs.Select(a => $"{a}/{TestConstants.TestMsiInstaller}));

and then down below you can assert that each expected arch is present in the Installers

palenshus (changes requested):
:clock1:

src/WingetCreateTests/WingetCreateTests/UnitTests/PackageParserTests.cs
ryfu-msft: Ah I didn't know I could do that, much cleaner now, changed to suggested.

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: Thanks for the suggestion. That is a much more exhaustive test. I did a little modification so now the test makes sure the architecture is parsed correctly.

ryfu-msft:
>         private bool VerifyUpdatedInstallerHash(Manifests oldManifest, InstallerManifest newManifest)

changed to static

---
In reply to: [881127753](https://github.com/microsoft/winget-create/pull/108#issuecomment-881127753) [](http://example.com/codeflow?ancestors=881127753,881127753)

---
Refers to: src/WingetCreateCLI/Commands/UpdateCommand.cs:334 in 1000920. [](commit_id = 10009203810225fff473a4ca146fa95b8ef021d7, deletion_comment = False)

palenshus (approved):
:shipit:

PR #112, opened 2021-07-16
Add settings configuration for specifying winget-pkgs repo owner/name

### LINKED ISSUE #95: Add configuration settings to specify owner and repo for winget-create
Add a configuration setting to the setting json file that allows user to specify the owner and repository that they want Winget-Create to interact with. This can be helpful for those who have their own private fork of winget-pkgs and want to make submissions there instead of on the main repository.

ryfu-msft:
This pull request resolve #95. You can now specify the wingetpkgsrepo owner and name to tell wingetcreate which repo to look for packages.

Changes:
- Added WingetPkgsRepoOwner and WingetPkgsRepoName to the settings schema.
- Added an repository error message not found if the repository is invalid.

Other Changes
- Changed  UpdateCommandTest (UpdateBasedOnInstallerUrlMatch) to use .Zip
- Moved DisplayMismatchedArchitecture to before error messages

Tests:
- Added settings command test to verify that the RepositoryNotFound error message is shown if the repo settings are invalid.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/112)


src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -116,7 +116,16 @@ public override async Task<bool> Execute()
 
                 Logger.DebugLocalized(nameof(Resources.RetrievingManifest_Message), this.Id);
 
-                string exactId = await client.FindPackageId(this.Id);
+                string exactId = string.Empty;
```
palenshus, line 119: Why empty instead of null?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -132,7 +132,9 @@ public override async Task<bool> Execute()
 
                 do
                 {
-                    if (!await this.PromptPackageIdentifierAndCheckDuplicates(manifests))
+                    if (this.WingetRepoOwner == DefaultWingetRepoOwner &&
```
palenshus, line 135: Why only for the main repo?

src/WingetCreateCLI/UserSettings.cs
```diff
@@ -53,6 +53,40 @@ public static bool TelemetryDisabled
             }
         }
 
+        /// <summary>
+        /// Gets or sets the owner of the winget-pkgs repository.
+        /// </summary>
+        public static string WingetPkgsRepoOwner
+        {
+            get
```
palenshus, line 61: Can simplify the gets to:

get => Settings.WingetPkgsRepo.Owner;

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: If a user is attempting to generate a new package manifest for their own test repo, we shouldn't block them from creating manifests or testing out manifest creation. I believe we should only apply the duplicate check if they're using the main repo.

palenshus:
>         public const string DefaultWingetRepoOwner = "microsoft";

Aren't these redundant with the defaults in the schema now?

---
Refers to: src/WingetCreateCLI/Commands/BaseCommand.cs:35 in 9a7c743. [](commit_id = 9a7c74367b5d42c3d0de75836e68d0595779fe9a, deletion_comment = False)

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: If a user is attempting to generate a new package manifest for their own test repo, we shouldn't block them from creating manifests or testing out manifest creation. I believe we should only apply the duplicate check if they're using the main repo.

palenshus: You're probably right, though this won't always be for "test" repos, eventually users may have their own alternate "prod" repos I believe, but anyway this is fine for now I think

ryfu-msft:
>         public const string DefaultWingetRepoOwner = "microsoft";

These are redundant with the defaults in the schema, but I believe these const strings still come in handy when checking against the main repo like for our testing in GitHubTestBase.cs which is why I left them there.

---
In reply to: [881777874](https://github.com/microsoft/winget-create/pull/112#issuecomment-881777874) [](http://example.com/codeflow?ancestors=881777874,881777874)

---
Refers to: src/WingetCreateCLI/Commands/BaseCommand.cs:35 in 9a7c743. [](commit_id = 9a7c74367b5d42c3d0de75836e68d0595779fe9a, deletion_comment = False)

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: changed to suggestion

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Oops, changed to uninitialized. I think I thought I was working with a var

palenshus:
>         public const string DefaultWingetRepoOwner = "microsoft";

If we wanted to get rid of the redundancy, we _could_ change these to:

        public static readonly string DefaultWingetRepoOwner = new WingetPkgsRepo().Owner;

No pressure though

---
In reply to: [881780347](https://github.com/microsoft/winget-create/pull/112#issuecomment-881780347) [](http://example.com/codeflow?ancestors=881780347,881777874)

---
Refers to: src/WingetCreateCLI/Commands/BaseCommand.cs:35 in 9a7c743. [](commit_id = 9a7c74367b5d42c3d0de75836e68d0595779fe9a, deletion_comment = False)

palenshus (approved):
:shipit:

PR #117, opened 2021-07-21
Modify telemetry first run experience to be non-blocking

ryfu-msft:
The first run experience waits for the user the provide their consent, which causes an issue for usage in a CI/CD scenario by creating a blocking event. To resolve this issue, we instead notify users of the default telemetry settings, but provide instructions as to how to turn them off. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/117)

### DIFF
--- a/src/WingetCreateCLI/UserSettings.cs
+++ b/src/WingetCreateCLI/UserSettings.cs
@@ -118,13 +118,13 @@ namespace Microsoft.WingetCreateCLI
         {
             if (!File.Exists(SettingsJsonPath))
             {
-                Prompt.Symbols.Done = new Symbol(string.Empty, string.Empty);
-                Prompt.Symbols.Prompt = new Symbol(string.Empty, string.Empty);
                 Console.WriteLine(Resources.TelemetrySettings_Message);
                 Console.WriteLine("------------------");
                 Console.WriteLine(Resources.TelemetryJustification_Message);
                 Console.WriteLine(Resources.TelemetryAnonymous_Message);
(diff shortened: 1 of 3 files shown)


src/WingetCreateCLI/UserSettings.cs
```diff
@@ -118,13 +117,13 @@ public static void FirstRunTelemetryConsent()
...
-                Prompt.Symbols.Done = new Symbol(string.Empty, string.Empty);
-                Prompt.Symbols.Prompt = new Symbol(string.Empty, string.Empty);
                 Console.WriteLine(Resources.TelemetrySettings_Message);
                 Console.WriteLine("------------------");
                 Console.WriteLine(Resources.TelemetryJustification_Message);
                 Console.WriteLine(Resources.TelemetryAnonymous_Message);
-                TelemetryDisabled = !Prompt.Confirm(Resources.EnableTelemetryFirstRun_Message);
+                Console.WriteLine(Resources.TelemetryEnabledByDefault_Message);
+                Console.WriteLine();
+                TelemetryDisabled = true;
```
palenshus, line 126: You're disabling telemetry at first run?

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/UserSettings.cs
ryfu-msft: Forgot to revert this, changed to false

palenshus (approved):

PR #119, opened 2021-07-21
Add cache command for managing downloaded installers

### LINKED ISSUE #62: wingetcreate storing installers in %temp% folder
### Description
When creating new or updating existing manifest, `wingetcreate` is given a download link to the installer so that it can download the application and generate sha256. I've noticed that `wingetcreate` stores the installer in `%temp%` folder after generating sha256. I like this behavior because if I had to do some testing with the installer or get some metadata from it, I would not have to download the installer again. But this could cause storage issue after many submissions so I would like to have a option to clear the installers via the application itself.

### Solution
add some command like `wingetcreate cache clean` to clear the installers.
(optional)
some extra commands like `wingetcreate cache --list` will display all the installers that wingetcreate downloaded. Doesn't have to be exact appname just the installer's name. Output could look similar to when we type `dir` command in a folder. This feature could be easily added if wingetcreate stored installers in it's own directory somewhere instead of `%temp%` folder so that it can just clone what `dir` command does.

ryfu-msft:
This pull request resolves #62  

Changes:
- Change installer download path from %temp% -> %temp%/wingetcreate
- Created a cache command that opens the cache folder, lists downloaded installers, and cleans out all downloaded installers
- Changed basecommand.githubtoken to be a virtual property that can be overrided so that commands like settings and cache don't inherit the --token option
- Removes GetVerbs() from program.cs as this was displaying commands in alphabetical order. Reverted to listing out the commands in an array to have an ordering based on usage.
- Added cache command documentation

Testing:
- Added unit tests for the cache command

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/119)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -44,11 +44,6 @@ namespace Microsoft.WingetCreateCLI.Commands
         protected const string ProgramApplicationAlias = "wingetcreate.exe";
 
         /// <summary>
-        /// Gets or sets the GitHub token used to submit a pull request on behalf of the user.
-        /// </summary>
-        public virtual string GitHubToken { get; set; }
-
-        /// <summary>
         /// Gets or sets the winget repo owner to use.
         /// </summary>
         public string WingetRepoOwner { get; set; } = UserSettings.WindowsPackageManagerRepositoryOwner ?? DefaultWingetRepoOwner;
@@ -77,97 +72,6 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// Gets the GitHubClient instance to use for interacting with GitHub from the CLI.
         /// </summary>
         protected GitHub GitHubClient { get; private set; }
-
-        /// <summary>
-        /// Validates the GitHubToken provided on the command-line, or if not present, the cached token if one exists.
-        /// Attempts a simple operation against the target repo, and if that fails, then:
-        ///     If token provided on command-line, errors out
-        ///     If not, and cached token was present, then deletes token cache, and starts OAuth flow
-        /// Otherwise, sets the instance variable to hold the validated token.
-        /// If no token is present on command-line or in cache, starts the OAuth flow to retrieve one.
-        /// </summary>
-        /// <param name="cacheToken">Boolean to override default behavior and force caching of token.</param>
-        /// <returns>True if the token is now present and valid, false otherwise.</returns>
-        public async Task<bool> SetAndCheckGitHubToken(bool cacheToken = false)
-        {
-            string cachedToken = null;
-            bool hasPatToken = !string.IsNullOrEmpty(this.GitHubToken);
-            string token = this.GitHubToken;
-
-            if (!hasPatToken)
-            {
-                Logger.Trace("No token parameter, reading cached token");
-                token = cachedToken = GitHubOAuth.ReadTokenCache();
-
-                if (string.IsNullOrEmpty(token))
-                {
-                    Logger.Trace("No cached token found.");
-                    Logger.DebugLocalized(nameof(Resources.GitHubAccountMustBeLinked_Message));
-                    Logger.DebugLocalized(nameof(Resources.ExecutionPaused_Message));
-                    Console.WriteLine();
-                    token = await GitHubOAuthLoginFlow();
-                    if (string.IsNullOrEmpty(token))
-                    {
-                        // User must've cancelled OAuth flow, we can't proceed successfully
-                        Logger.WarnLocalized(nameof(Resources.NoTokenResponse_Message));
-                        return false;
-                    }
-
-                    Logger.DebugLocalized(nameof(Resources.ResumingCommandExecution_Message));
-                }
-                else
-                {
-                    Logger.DebugLocalized(nameof(Resources.UsingTokenFromCache_Message));
-                }
-            }
-
-            this.GitHubClient = new GitHub(token, this.WingetRepoOwner, this.WingetRepo);
-
-            try
-            {
-                Logger.Trace("Checking repo access using OAuth token");
-                await this.GitHubClient.CheckAccess();
-                Logger.Trace("Access check was successful, proceeding");
-                this.GitHubToken = token;
-
-                // Only cache the token if it came from Oauth, instead of PAT parameter or cache
-                if (cacheToken || (!hasPatToken && token != cachedToken))
-                {
-                    try
-                    {
-                        Logger.Trace("Writing token to cache");
-                        GitHubOAuth.WriteTokenCache(token);
-                    }
-                    catch (Exception ex)
-                    {
-                        // Failing to cache the token shouldn't be fatal.
-                        Logger.WarnLocalized(nameof(Resources.WritingCacheTokenFailed_Message), ex.Message);
-                    }
-                }
-
-                return true;
-            }
-            catch (Exception e)
-            {
-                if (token == cachedToken)
-                {
-                    // There's an issue with the cached token, so let's delete it and try again
-                    Logger.WarnLocalized(nameof(Resources.InvalidCachedToken));
-                    GitHubOAuth.DeleteTokenCache();
-                    return await this.SetAndCheckGitHubToken();
-                }
-                else if (e is AuthorizationException)
-                {
-                    Logger.ErrorLocalized(nameof(Resources.Error_Prefix), e.Message);
-                    Logger.ErrorLocalized(nameof(Resources.InvalidTokenError_Message));
-                    return false;
-                }
-                else
-                {
-                    throw;
-                }
-            }
-        }
 
         /// <summary>
         /// Abstract method executing the command.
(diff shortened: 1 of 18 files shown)


src/WingetCreateCLI/Program.cs
```diff
@@ -38,7 +38,7 @@ private static async Task<int> Main(string[] args)
 
             Parser myParser = new Parser(config => config.HelpWriter = null);
 
-            var types = GetVerbs();
+            var types = new Type[] { typeof(NewCommand), typeof(UpdateCommand), typeof(SubmitCommand), typeof(SettingsCommand), typeof(CacheCommand) };
```
palenshus, line 41: Nice

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <summary>
+    /// Command to manage downloaded installers found in the %TEMP%/wingetcreate folder.
+    /// </summary>
+    [Verb("cache", HelpText = "CacheCommand_HelpText", ResourceType = typeof(Resources))]
+    public class CacheCommand : BaseCommand
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to delete all downloaded installers found in cached.
+        /// </summary>
+        [Option('c', "clean", SetName = nameof(Clean), Required = true, HelpText = "Clean_HelpText", ResourceType = typeof(Resources))]
```
palenshus, line 26: Why are these all Required:true?

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    Directory.CreateDirectory(PackageParser.InstallerDownloadPath);
+                }
+
+                if (this.Clean)
+                {
+                    Logger.InfoLocalized(nameof(Resources.CleaningInstallers_Message), PackageParser.InstallerDownloadPath);
+                    DirectoryInfo dir = new DirectoryInfo(PackageParser.InstallerDownloadPath);
+
+                    foreach (FileInfo file in dir.GetFiles())
```
palenshus, line 64: Always use Enumerate* unless you have a good reason not to

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                    {
+                        file.Delete();
+                    }
+
+                    commandEvent.IsSuccessful = true;
+                }
+                else if (this.List)
+                {
+                    string[] files = Directory.GetFiles(PackageParser.InstallerDownloadPath);
+                    Logger.InfoLocalized(nameof(Resources.InstallersFound_Message), files.Length, PackageParser.InstallerDownloadPath);
```
palenshus, line 74: I'd display this for Clean too

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                    Console.WriteLine();
+
+                    foreach (string file in files)
+                    {
+                        Logger.Info(Path.GetFileName(file));
```
palenshus, line 79: Probably a good idea to do this for Clean too

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    Logger.InfoLocalized(nameof(Resources.CleaningInstallers_Message), PackageParser.InstallerDownloadPath);
+                    DirectoryInfo dir = new DirectoryInfo(PackageParser.InstallerDownloadPath);
+
+                    foreach (FileInfo file in dir.GetFiles())
+                    {
+                        file.Delete();
+                    }
+
+                    commandEvent.IsSuccessful = true;
```
palenshus, line 69: This should be right before the return statemetn

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -47,8 +46,7 @@ public abstract class BaseCommand
         /// <summary>
         /// Gets or sets the GitHub token used to submit a pull request on behalf of the user.
         /// </summary>
-        [Option('t', "token", Required = false, HelpText = "GitHubToken_HelpText", ResourceType = typeof(Resources))]
-        public string GitHubToken { get; set; }
+        public virtual string GitHubToken { get; set; }
```
palenshus, line 49: Is there any point to having this on the base class at all anymore?

src/WingetCreateTests/WingetCreateTests/UnitTests/CacheCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// Verifies that if the Clean flag is present, deletes all downloaded installers present in the cache folder.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task CleanCachedInstallers()
+        {
+            CacheCommand command = new CacheCommand() { Clean = true };
+            await command.Execute();
+            var installerFiles = Directory.GetFiles(PackageParser.InstallerDownloadPath);
+            Assert.IsTrue(installerFiles.Length == 0, "Cached installers were not deleted.");
```
palenshus, line 67: Whenever comparing a number, do a Assert.AreEqual instead of IsTrue

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Program.cs
ryfu-msft: Thanks :)

src/WingetCreateCLI/Commands/CacheCommand.cs
ryfu-msft: Similar to the token command, I have to make these all required or else someone could run 'wingetcreate cache' and no parsing errors would appear. Another alternative would be to check if these values are set and then show an error message, but I would want parsing errors to be handled by the commandline parser and not by the command

src/WingetCreateCLI/Commands/CacheCommand.cs
ryfu-msft: Added additional message for clean

src/WingetCreateCLI/Commands/CacheCommand.cs
ryfu-msft: Added additional message for clean

src/WingetCreateCLI/Commands/CacheCommand.cs
ryfu-msft: Moved boolean assignment to return statement

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I think you're right. Only the SetAndCheckGitHubToken method uses this property, which we can easily make into a parameter. Removed this virtual property from the base class

src/WingetCreateTests/WingetCreateTests/UnitTests/CacheCommandTests.cs
ryfu-msft: Oops, I knew that! Fixed

src/WingetCreateCLI/Program.cs
```diff
@@ -95,48 +88,20 @@ private static void DisplayHelp(NotParsed<object> result)
...
-                {
-                    Utils.WriteLineColored(ConsoleColor.Red, $"Command line parsing error: {error.Tag}");
-                }
-            }
-        }
-    }
+            Console.WriteLine();
+        }
+
+        private static void DisplayParsingErrors<T>(ParserResult<T> result)
```
palenshus, line 94: Wow, this sure looks a lot cleaner! Looks like excList is never used though?

palenshus:
Hit an unrelated error where if the AppStatePath wasn't already created at first run, the program would crash. Recommend creating it just to be sure:

        private static readonly Lazy<string> AppStatePathLazy = new(() =>
        {
            string path = IsRunningAsUwp()
                ? ApplicationData.Current.LocalFolder.Path
                : Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Microsoft", ModuleName);
            Directory.CreateDirectory(path);
            return path;
        });

src/WingetCreateTests/WingetCreateTests/UnitTests/CacheCommandTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// Verifies that if the Clean flag is present, deletes all downloaded installers present in the cache folder.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Test]
+        public async Task CleanCachedInstallers()
+        {
+            CacheCommand command = new CacheCommand() { Clean = true };
+            await command.Execute();
+            var installerFiles = Directory.GetFiles(PackageParser.InstallerDownloadPath);
+            Assert.AreEqual(0, installerFiles.Length, "Cached installers were not deleted.");
```
palenshus, line 67: Can the UT change InstallerDownloadPath temporarily (in setup/teardown) so it doesn't mess with the existing installer cache?

src/WingetCreateCLI/Commands/BaseCommand.cs
palenshus: Oops, that won't work, since that method actually sets this.GitHubToken internally. If you just change it to a parameter it won't change the value

palenshus:
```
Launching web browser... If default web browser does not launch, please go to: [https://github.com/login/device](https://github.com/login/device)
Please enter the following user verification code when prompted: CAAD-3153

Resuming command execution...
No token provided, submission to GitHub skipped.
```

src/WingetCreateCLI/Commands/CacheCommand.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+                else if (this.List)
+                {
+                    string[] files = Directory.GetFiles(PackageParser.InstallerDownloadPath);
+                    Logger.InfoLocalized(nameof(Resources.InstallersFound_Message), files.Length, PackageParser.InstallerDownloadPath);
+                    Console.WriteLine();
+
+                    foreach (string file in files)
+                    {
+                        Logger.Info(Path.GetFileName(file));
```
palenshus, line 83: IMO, this should be Logger.Debug. It looks weird to me to see every line in green

ryfu-msft:
Changed to suggestion

---
In reply to: [885148402](https://github.com/microsoft/winget-create/pull/119#issuecomment-885148402) [](http://example.com/codeflow?ancestors=885148402,885148402)

src/WingetCreateCLI/Program.cs
ryfu-msft: removed excList

src/WingetCreateCLI/Commands/CacheCommand.cs
ryfu-msft: Changed to Logger.Debug

src/WingetCreateTests/WingetCreateTests/UnitTests/CacheCommandTests.cs
ryfu-msft: Added a way to modify the InstallerDownloadPath so that the UT can change it to a different path that wont' interfere with existing installer cache

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Ahh, thanks for catching my mistake, I reverted it back to what we had before. I think we'll need to keep this around for now. I'll look into this further when I refactor some of the oauth stuff in a future PR.

src/WingetCreateCLI/Commands/CacheCommand.cs
palenshus: It does result in some confusing output, but unfortunately CommandLine doesn't have a way to do mutually exclusive option sets :( I guess this is okay, if not ideal

palenshus: This issue tracks enabling this [https://github.com/commandlineparser/commandline/issues/419](https://github.com/commandlineparser/commandline/issues/419)

src/WingetCreateTests/WingetCreateTests/UnitTests/CacheCommandTests.cs
palenshus: Lovely

palenshus (approved):
:shipit:

Issue #116, opened 2021-07-21
Cached Github token not being used for non-submission API calls, leading to rate limits


apetresc:
# Brief description of your issue

During normal use of `wingetcreate.exe`, I will very quickly run into the dreaded:
```
Octokit.RateLimitExceededException: API rate limit exceeded for 99.250.197.152. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
   at Octokit.Connection.HandleErrors(IResponse response) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/Connection.cs:line 700
   at Octokit.Connection.RunRequest(IRequest request, CancellationToken cancellationToken) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/Connection.cs:line 681
   at Octokit.Connection.Run[T](IRequest request, CancellationToken cancellationToken) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/Connection.cs:line 666
   at Octokit.ApiConnection.GetPage[TU](Uri uri, IDictionary`2 parameters, String accepts, ApiOptions options) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/ApiConnection.cs:line 628
   at Octokit.ApiConnection.<>c__DisplayClass18_0`1.<<GetAll>b__0>d.MoveNext() in /home/runner/work/octokit.net/octokit.net/Octokit/Http/ApiConnection.cs:line 212
--- End of stack trace from previous location ---
   at Octokit.ApiPagination.GetAllPages[T](Func`1 getFirstPage, Uri uri)
   at Microsoft.WingetCreateCore.Common.GitHub.FindPackageIdRecursive(String[] packageId, String path, String exactPackageId, Int32 index)
   at Microsoft.WingetCreateCore.Common.GitHub.FindPackageId(String packageId)
   at Microsoft.WingetCreateCLI.Commands.UpdateCommand.Execute()
   at Microsoft.WingetCreateCLI.Program.Main(String[] args)
{"message":"API rate limit exceeded for 99.250.197.152. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```
This is just after maybe ~10-20 calls to `wingetcreate.exe update [...]`.

[...]

PR #121, opened 2021-07-23
Notify users of latest release version

### LINKED ISSUE #66: Check at startup to determine if there's a newer release and prompt user to update

We should check GitHub at startup to see if there's a newer release available than what the user is currently running, and if so, prompt the user to update (maybe provide the winget command they'd run to update for ease)

(Potentially fail execution if the newest release is a major version update, to prevent user creating broken manifests?)

Call [GitHub releases API](https://docs.github.com/en/rest/reference/repos#releases) at startup, and check against assembly version, acting with above logic

ryfu-msft:
This PR resolves #66.

Changes:
- Added a static method in GitHub.cs that retrieves the tag name of the latest release version.
- If the current version of winget-create does not match the latest version, then a warning message appears notifying the user that an update is available.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/121)

### DIFF
--- a/src/WingetCreateCLI/Program.cs
+++ b/src/WingetCreateCLI/Program.cs
@@ -35,12 +35,16 @@ namespace Microsoft.WingetCreateCLI
             var parserResult = myParser.ParseArguments(args, types);
 
             BaseCommand command = parserResult.MapResult(c => c as BaseCommand, err => null);
+            if (!await command.LoadGitHubClient())
+            {
+                return 1;
+            }
 
-            string token = command?.GitHubToken ?? GitHubOAuth.ReadTokenCache();
             try
             {
-                string latestVersion = await GitHub.GetLatestRelease(token);
-                if (latestVersion.TrimStart('v').Split('-').First() != Utils.GetEntryAssemblyVersion())
+                string latestVersion = await GitHub.GetLatestRelease();
+                string trimmedVersion = latestVersion.TrimStart('v').Split('-').First();
+                if (trimmedVersion != Utils.GetEntryAssemblyVersion())
                 {
                     Logger.WarnLocalized(nameof(Resources.OutdatedVersionNotice_Message));
                     Logger.WarnLocalized(nameof(Resources.GetLatestVersion_Message), latestVersion, "https://github.com/microsoft/winget-create/releases");
@@ -50,6 +54,7 @@ namespace Microsoft.WingetCreateCLI
             }
             catch (Exception ex) when (ex is Octokit.ApiException || ex is Octokit.RateLimitExceededException)
             {
+                // Since this is only notifying the user if an update is available, don't block if the token is invalid or a rate limit error is encountered.
             }
 
             if (command == null)
--- a/src/WingetCreateCLI/Program.cs
+++ b/src/WingetCreateCLI/Program.cs
@@ -26,6 +26,14 @@ namespace Microsoft.WingetCreateCLI
             UserSettings.FirstRunTelemetryConsent();
             TelemetryEventListener.EventListener.IsTelemetryEnabled();
 
+            string release = await GitHub.GetLatestRelease();
+            if (!release.Contains(Utils.GetEntryAssemblyVersion()))
+            {
+                Logger.WarnLocalized(nameof(Resources.OutdatedVersionNotice_Message));
+                Logger.WarnLocalized(nameof(Resources.GetLatestVersion_Message), release, "https://github.com/microsoft/winget-create/releases");
+                Console.WriteLine();
+            }
+
             string arguments = string.Join(' ', Environment.GetCommandLineArgs());
             Logger.Trace($"Command line args: {arguments}");
 
@@ -89,6 +97,7 @@ namespace Microsoft.WingetCreateCLI
         private static void DisplayParsingErrors<T>(ParserResult<T> result)
         {
             var builder = SentenceBuilder.Create();
+
             var errorMessages = HelpText.RenderParsingErrorsTextAsLines(result, builder.FormatError, builder.FormatMutuallyExclusiveSetErrors, 1);
 
             foreach (var error in errorMessages)
--- a/src/WingetCreateCLI/Program.cs
+++ b/src/WingetCreateCLI/Program.cs
@@ -37,22 +37,19 @@ namespace Microsoft.WingetCreateCLI
             BaseCommand command = parserResult.MapResult(c => c as BaseCommand, err => null);
 
             string token = command?.GitHubToken ?? GitHubOAuth.ReadTokenCache();
-            if (!string.IsNullOrEmpty(token))
+            try
             {
-                try
+                string latestVersion = await GitHub.GetLatestRelease(token);
+                if (latestVersion.TrimStart('v').Split('-').First() != Utils.GetEntryAssemblyVersion())
                 {
-                    string latestVersion = await GitHub.GetLatestRelease(token);
-                    if (latestVersion.TrimStart('v').Split('-').First() != Utils.GetEntryAssemblyVersion())
-                    {
-                        Logger.WarnLocalized(nameof(Resources.OutdatedVersionNotice_Message));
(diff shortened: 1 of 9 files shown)


src/WingetCreateCLI/Program.cs
```diff
@@ -25,6 +25,14 @@ private static async Task<int> Main(string[] args)
             Logger.Initialize();
             UserSettings.FirstRunTelemetryConsent();
             TelemetryEventListener.EventListener.IsTelemetryEnabled();
+
+            string latestVersion = await GitHub.GetLatestRelease();
+            if (!latestVersion.Contains(Utils.GetEntryAssemblyVersion()))
```
palenshus, line 30: This might be a little too fuzzy, for example:

v11.0.0.0-preview would contain 1.0.0.0

How about:

(latestVersion.TrimStart('v').Split('-').First() == Utils.GetEntryAssemblyVersion() instead?

src/WingetCreateCLI/Program.cs
```diff
@@ -25,6 +25,14 @@ private static async Task<int> Main(string[] args)
             Logger.Initialize();
             UserSettings.FirstRunTelemetryConsent();
             TelemetryEventListener.EventListener.IsTelemetryEnabled();
+
+            string latestVersion = await GitHub.GetLatestRelease();
+            if (!latestVersion.Contains(Utils.GetEntryAssemblyVersion()))
+            {
+                Logger.WarnLocalized(nameof(Resources.OutdatedVersionNotice_Message));
+                Logger.WarnLocalized(nameof(Resources.GetLatestVersion_Message), latestVersion, "https://github.com/microsoft/winget-create/releases");
```
palenshus, line 33: @denelon should we include a winget commandline too? 

winget upgrade Microsoft.WingetCreate

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -59,6 +59,18 @@ public static async Task<string> GetGitHubAppInstallationAccessToken(string gitH
...
+        }
+
+        /// <summary>
+        /// Gets the latest release tag name of winget-create.
+        /// </summary>
+        /// <returns>Latest release tag name.</returns>
+        public static async Task<string> GetLatestRelease()
+        {
+            var github = new GitHubClient(new ProductHeaderValue(UserAgentName));
+            var releases = await github.Repository.Release.GetAll("microsoft", "winget-create");
```
palenshus, line 71: Why not use GetLatest? Should be faster too

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -59,6 +59,18 @@ public static async Task<string> GetGitHubAppInstallationAccessToken(string gitH
...
             return response.Token;
+        }
+
+        /// <summary>
+        /// Gets the latest release tag name of winget-create.
+        /// </summary>
+        /// <returns>Latest release tag name.</returns>
+        public static async Task<string> GetLatestRelease()
+        {
+            var github = new GitHubClient(new ProductHeaderValue(UserAgentName));
```
palenshus, line 70: This is going to hit [https://github.com/microsoft/winget-create/issues/116](https://github.com/microsoft/winget-create/issues/116), potentially at the beginning of every launch. This is tricky, since this may run for a command that doesn't even have the token parameter.

I'd say at minimum use the user's token parameter if it exists, and when you do your refactor, use the cached token as well. 

OR

If this fails due to rate limiting, just ignore that failure

palenshus (changes requested):
:clock1:

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: I hadn't thought about rate limits when checking for release. I will probably hold off on this PR until the token caching refactor is complete.

src/WingetCreateCLI/Program.cs
ryfu-msft: Changed to suggestion, thanks for catching that mistake, can't wait for v11 of wingetcreate!

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: Changed to using GetLatest

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: I have changed the logic to be similar to what you suggested. The way it works now is that if a token is present (either as an argument from cache) it will use that to get the latest version. If not, it will still try to get the latest version, and any rate limiting failures are caught and ignored.

src/WingetCreateCLI/Program.cs
ryfu-msft: Added the winget upgrade to the warning message

src/WingetCreateCLI/Program.cs
```diff
@@ -1,88 +1,104 @@
...
+
+            Parser myParser = new Parser(config => config.HelpWriter = null);
 
             var types = new Type[] { typeof(NewCommand), typeof(UpdateCommand), typeof(SubmitCommand), typeof(SettingsCommand), typeof(TokenCommand), typeof(CacheCommand) };
             var parserResult = myParser.ParseArguments(args, types);
 
             BaseCommand command = parserResult.MapResult(c => c as BaseCommand, err => null);
 
-            if (command == null)
+            string token = command?.GitHubToken ?? GitHubOAuth.ReadTokenCache();
```
palenshus, line 39: Why not just call command.LoadGitHubClient() and then use that in GetLatestRelease instead of passing in a token?

src/WingetCreateCLI/Program.cs
```diff
@@ -1,88 +1,104 @@
...
+            try
+            {
+                string latestVersion = await GitHub.GetLatestRelease(token);
+                if (latestVersion.TrimStart('v').Split('-').First() != Utils.GetEntryAssemblyVersion())
```
palenshus, line 43: Just for clarity I'd make a new var for the trimmedVersion

src/WingetCreateCLI/Program.cs
```diff
@@ -1,88 +1,104 @@
...
+                {
+                    Logger.WarnLocalized(nameof(Resources.OutdatedVersionNotice_Message));
+                    Logger.WarnLocalized(nameof(Resources.GetLatestVersion_Message), latestVersion, "https://github.com/microsoft/winget-create/releases");
+                    Logger.WarnLocalized(nameof(Resources.UpgradeUsingWinget_Message));
+                    Console.WriteLine();
+                }
+            }
+            catch (Exception ex) when (ex is Octokit.ApiException || ex is Octokit.RateLimitExceededException)
+            {
```
palenshus, line 52: Add a comment in the block to explain why we're ignoring the exception

ryfu-msft: Added comment to block explaining why we're ignoring exception

src/WingetCreateCLI/Program.cs
ryfu-msft: created new variable for clarity

src/WingetCreateCLI/Program.cs
ryfu-msft: I like this idea, but it will require some changes like making LoadGitHubClient() a public method as well as storing the token as a private static string in the GitHub.cs class in order to keep GetLatestRelease static.

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -62,6 +64,23 @@ public static async Task<string> GetGitHubAppInstallationAccessToken(string gitH
             return response.Token;
         }
 
+        /// <summary>
+        /// Gets the latest release tag name of winget-create.
+        /// </summary>
+        /// <returns>Latest release tag name.</returns>
+        public static async Task<string> GetLatestRelease()
```
palenshus, line 71: You can make this non-static now that you have a githubclient in Program.cs

ryfu-msft: Changed method to non-static

palenshus (approved):
:shipit:

PR #125, opened 2021-07-23
Refactor SetAndCheckGitHubToken method and handle rate limit throttling

### LINKED ISSUE #116: Cached Github token not being used for non-submission API calls, leading to rate limits

ryfu-msft:
This PR fixes #116.

Changes:
The SetAndCheckGitHubToken method aimed to handle too much like reading from cache, checking access, storing and deleting cache token, and initiating oauth login. 

Because of this, I decided to refactor the method into smaller ones so that the smaller components can be easier to utilize. This includes checking the cache for a stored token to help prevent rate limit throttling when using the update command without providing a token.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/125)


src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,143 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
-                    Logger.DebugLocalized(nameof(Resources.ExecutionPaused_Message));
-                    Console.WriteLine();
-                    token = await GitHubOAuthLoginFlow();
-                    if (string.IsNullOrEmpty(token))
-                    {
-                        // User must've cancelled OAuth flow, we can't proceed successfully
-                        Logger.WarnLocalized(nameof(Resources.NoTokenResponse_Message));
-                        return false;
-                    }
+                    Logger.WarnLocalized(nameof(Resources.RepeatedUseWarning_Message));
```
palenshus, line 327: IMO preemptively warning about this seems overly pedantic to me. 95% of users will probably never encounter this

palenshus:
> // Copyright (c) Microsoft. All rights reserved.

Revert this file, only whitespace changed

---
Refers to: src/WingetCreateCore/Common/GitHub.cs:1 in 58bf882. [](commit_id = 58bf882c4355b72f64c33cdf91f0c3e34b8d58ac, deletion_comment = False)

src/WingetCreateCLI/Commands/BaseCommand.cs
palenshus: Also it'd be strange that this only happens for the update command

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,143 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
+                    Logger.WarnLocalized(nameof(Resources.AvoidRateLimitWarning_Message));
+                }
+            }
+            else
+            {
+                await this.StoreTokenInCache();
```
palenshus, line 333: This is a behavior change, before we only cached the token from oauth, not from the -t parameter. I can't remember why we decided to do that exactly, but we should make sure we're intentionally changing that

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -379,12 +379,9 @@ private async Task<bool> PromptPackageIdentifierAndCheckDuplicates(Manifests man
...
             GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
 
-            if (!string.IsNullOrEmpty(this.GitHubToken))
+            if (string.IsNullOrEmpty(this.GitHubToken))
             {
-                if (!await this.SetAndCheckGitHubToken())
-                {
-                    return false;
-                }
+                this.GitHubToken = await this.GetTokenFromCache();
```
palenshus, line 384: Doesn't make sense to do this after already creating the client?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -379,12 +379,9 @@ private async Task<bool> PromptPackageIdentifierAndCheckDuplicates(Manifests man
         {
             GitHub client = new GitHub(this.GitHubToken, this.WingetRepoOwner, this.WingetRepo);
```
palenshus, line 380: Shouldn't this call CheckForGitHubToken too? This seems to be the same circumstance as in UpdateCommand.cs

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -127,9 +121,17 @@ public override async Task<bool> Execute()
...
                 }
-                catch (Octokit.NotFoundException)
+                catch (Exception e)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.RepositoryNotFound_Error), this.WingetRepoOwner, this.WingetRepo);
+                    if (e is Octokit.NotFoundException)
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.RepositoryNotFound_Error), this.WingetRepoOwner, this.WingetRepo);
+                    }
+                    else if (e is Octokit.RateLimitExceededException)
```
palenshus, line 130: Why rate limit check here, but not elsewhere, like in NewCommand.cs?

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,143 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
-        ///     If token provided on command-line, errors out
-        ///     If not, and cached token was present, then deletes token cache, and starts OAuth flow
-        /// Otherwise, sets the instance variable to hold the validated token.
-        /// If no token is present on command-line or in cache, starts the OAuth flow to retrieve one.
+        /// If no GitHub token is provided, looks in the token cache and assigns it.
         /// </summary>
-        /// <param name="cacheToken">Boolean to override default behavior and force caching of token.</param>
-        /// <returns>True if the token is now present and valid, false otherwise.</returns>
-        protected async Task<bool> SetAndCheckGitHubToken(bool cacheToken = false)
+        protected async void CheckForGitHubToken()
```
palenshus, line 318: I wonder if it would make sense to instead of exposing all the methods, expose just one method, something like 

```
public GitHub GitHubGetGitHubClient(bool requireToken)
{
    // Create GitHub client, check access, rate-limiting, etc. If CheckAccess get a rate-limit error, it could automatically do the Oauth flow too

    // If requireToken set, then retrieve token from this.Token, or from cache, and check that, otherwise get from oauth
}
```

That way everywhere in the code that needs to interact with GitHub can just call this method to get a client, and it'll do all the logic internally to make sure the client is as valid as needed for this scenario.

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I think I agree with you. I removed these warnings so now it only shows when the error is hit.

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I am aware of this behavior change. I probably should have run it by you first to be sure. Are you okay with this change? If the -t <token> they use is valid, I think it should be safe to cache it for the user for future use. Let me know if you are/aren't okay with that

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Removed, no longer needed.

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I like this idea because it keeps the client consistent for all commands. I implemented something similar to what you suggested called LoadGitHubClient. As long as the commands call this method to initialize the client, any further interaction with the client should be pretty seamless. let me know what you think.

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: No longer needed anymore, see LoadGitHubClient() method in base command class.

ryfu-msft:
> // Copyright (c) Microsoft. All rights reserved.

Reverted this file.

---
In reply to: [887075907](https://github.com/microsoft/winget-create/pull/125#issuecomment-887075907) [](http://example.com/codeflow?ancestors=887075907,887075907)

---
Refers to: src/WingetCreateCore/Common/GitHub.cs:1 in 58bf882. [](commit_id = 58bf882c4355b72f64c33cdf91f0c3e34b8d58ac, deletion_comment = False)

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Moved rate limit checking to the LoadGitHubClient method.

src/WingetCreateCLI/Commands/BaseCommand.cs
palenshus: I'm fine with it, adding @kevinla and @denelon to make sure they are too

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,133 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
 
         /// <summary>
-        /// Validates the GitHubToken provided on the command-line, or if not present, the cached token if one exists.
-        /// Attempts a simple operation against the target repo, and if that fails, then:
-        ///     If token provided on command-line, errors out
-        ///     If not, and cached token was present, then deletes token cache, and starts OAuth flow
-        /// Otherwise, sets the instance variable to hold the validated token.
-        /// If no token is present on command-line or in cache, starts the OAuth flow to retrieve one.
+        /// Creates a new GitHub client using the provided or cached token if present.
+        /// If the requireToken bool is set to TRUE, then the OAuth flow is launched to acquire a new token for the client.
```
palenshus, line 317: Clarify that Oauth is only launched if there's no token provided in command line or in cache

src/WingetCreateCLI/Commands/BaseCommand.cs
palenshus: Do we want to automatically launch the oauth flow if we're rate limited?

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,133 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
-                    Logger.ErrorLocalized(nameof(Resources.Error_Prefix), e.Message);
-                    Logger.ErrorLocalized(nameof(Resources.InvalidTokenError_Message));
-                    return false;
+                    Logger.WarnLocalized(nameof(Resources.InvalidGitHubToken_Message));
                 }
-                else
+                else if (e is RateLimitExceededException)
                 {
-                    throw;
+                    Logger.WarnLocalized(nameof(Resources.RateLimitExceeded_Message));
```
palenshus, line 431: Do we want to automatically launch the oauth flow if we're rate limited?

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Added additional clarification message

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Talked offline. I think we shouldn't as a clear error message with instructions on how to bypass this error should be sufficient and user friendly

src/WingetCreateCLI/Commands/BaseCommand.cs
denelon: I don't have any issues keeping that token for future use. We could address it in the future if anyone has a complaint. I don't think there are too many cases where people would be juggling different GitHub identities.

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -658,4 +649,19 @@
...
     <value>Installer cache cleaned.</value>
   </data>
+  <data name="DeletingInvalidCachedToken_Message" xml:space="preserve">
+    <value>Cached token was invalid, deleting token from cache...</value>
+  </data>
+  <data name="InvalidGitHubToken_Message" xml:space="preserve">
+    <value>Token was invalid. Please generate a new GitHub token and try again.</value>
+  </data>
+  <data name="RateLimitExceeded_Message" xml:space="preserve">
+    <value>GitHub api rate limit exceeded. To extend your rate limit, provide your GitHub token with the "-t" flag or store one using the "token --store" command.</value>
```
palenshus, line 659: I'm not sure if the localizers require you to explicitly state that parameters like this shouldn't be localized?

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -313,94 +313,134 @@ protected static void DisplayManifestPreview(Manifests manifests)
...
         /// <summary>
-        /// Validates the GitHubToken provided on the command-line, or if not present, the cached token if one exists.
-        /// Attempts a simple operation against the target repo, and if that fails, then:
-        ///     If token provided on command-line, errors out
-        ///     If not, and cached token was present, then deletes token cache, and starts OAuth flow
-        /// Otherwise, sets the instance variable to hold the validated token.
-        /// If no token is present on command-line or in cache, starts the OAuth flow to retrieve one.
+        /// Creates a new GitHub client using the provided or cached token if present.
+        /// If the requireToken bool is set to TRUE, OAuth flow can be launched to acquire a new token for the client.
+        /// The OAuth flow will only be launched if no token is provided in the command line or no token is present in the token cache.
```
palenshus, line 318: nit: "or" should be "and" here

src/WingetCreateCLI/Properties/Resources.resx
ryfu-msft: Added explicit comment explaining that these are parameters

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Changed to "and"

palenshus (approved):

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: @KevinLaMS  and I talked offline. We decided it would be better to not cache the token automatically when provided as a -t token as this could be a behavior that the user doesn't want/isn't expecting. The token command should be sufficient for storing tokens and this PR fixes the bug where other commands weren't consuming the cached token when calling github apis. I have reverted this change of behavior.

PR #126, opened 2021-07-24
Ensure retrieved files from github can only be ".yaml"

### LINKED ISSUE #124: WingetCreate copies all files.  It should be limited to .YAML
When updating a package, wingetcreate will pull down the previous version (yay!) so that most fields are already filled in.

The problem is that it pulls down more than *.yaml.  Since only YAML files can be submitted, this can cause submission errors.  Please only recreate YAML files.

ryfu-msft:
Fixes #124 

Adds an additional check when retrieving files from GitHub that the extension must be == ".yaml".

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/126)


src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -103,7 +104,7 @@ public async Task<List<string>> GetLatestManifestContentAsync(string packageId)
                 .FirstOrDefault();
 
             var packageContents = (await this.github.Repository.Content.GetAllContents(this.wingetRepoOwner, this.wingetRepo, version))
-                .Where(c => c.Type != ContentType.Dir);
+                .Where(c => c.Type != ContentType.Dir && Path.GetExtension(c.Name) == ".yaml");
```
palenshus, line 107: should be case-insensitive

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -1,142 +1,142 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license.
-
-namespace Microsoft.WingetCreateCLI.Commands
-{
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.WingetCreateCLI.Commands
```
palenshus, line 4: Revert this file, only whitespace changed

src/WingetCreateCore/Common/GitHub.cs
ryfu-msft: Changed to a case-insensitive string comparison

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: reverted this file

src/WingetCreateCore/Common/GitHub.cs
palenshus: We have an extension method for this already, EqualsIC

palenshus (approved):

PR #133, opened 2021-08-12
Add Navigational Menu to Edit Optional Installer Fields

### LINKED ISSUE #37: wingetcreate should prompt for optional fields.

The Windows Package Manager Manifest Creator should prompt users for optional fields like "InstallerSuccessCodes".

https://github.com/microsoft/winget-pkgs/issues/14224

After the required fields for a "valid" manifest has been created, the user should be prompted if they would like to add optional fields. If this is an "update" command, the current values should be displayed so they can be validated/modified.

ryfu-msft:
This pull request completes the optional feature and resolves #37 

Changes:
- Adds a navigational menu to allow users to either select all installers, or an individual installer to edit.
- Once the installer(s) is selected, a selection menu will be shown to the user listing out all optional installer fields for the user to select and edit.
- The field list will change depending on the installertype of the installer.
- Certain fields are automatically restricted from being edited by the user by default (InstallerURL, InstallerSha256, SignatureSha256, ProductCode (unless MSI), and PackageFamilyName(unless MSIX))
- Refactor the prompting methods into a new class called PromptHelper.cs 
- PromptAndSetPropertyValue() method is the main prompting method and should now be able to handle any field regardless of the field type.


src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+        }
+
+        private static List<string> GenerateInstallerSelectionList(List<Installer> installers)
+        {
+            List<string> selectionList = new List<string>();
+            selectionList.Add(Resources.AllInstallers_MenuItem);
+
+            foreach (Installer installer in installers)
+            {
+                var installerTuple = string.Join('|', installer.Architecture, installer.InstallerType.ToEnumAttributeValue(), installer.InstallerUrl);
```
palenshus, line 335: IMO this would look better with a space on either side of the pipe, " | "

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+                if (installer.Scope != null)
```
palenshus, line 336: These should show up before the url also

src/WingetCreateCLI/Commands/NewCommand.cs
palenshus, line 335: Could potentially simplify this all to:

string.Join(" | ", new[] { arch, type, scope, locale, url }.Where(t => t != null))

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -0,0 +1,270 @@
...
+                .Select(property => property.Name)
+                .Where(pName => !NonEditableOptionalFields.Any(field => field == pName))
+                .Append(exitMenuWord)
+                .ToList();
+
+            var installerTypeProperty = model.GetType().GetProperty(nameof(InstallerType));
+            if (installerTypeProperty != null)
+            {
+                var installerType = installerTypeProperty.GetValue(model);
+                if (installerType != null)
```
palenshus, line 65: There was more than just these two, see full comment in spec:

James would know better than me, but I'm not sure if Scope applies to msix? Can you have a machine-scope msix?

Can Platform apply to anything but MSIX (isn't everything else just "Desktop"?) PackageFamilyName is MSIX-only.

A few things don't apply to MSIX since they don't have install command lines. Things like Switches, success codes, maybe also UpgradeBehavior? InstallModes?

ProductCode is MSI-only (or "exe", since it may have an embedded MSI)

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -0,0 +1,270 @@
...
+            nameof(Installer.AdditionalProperties),
+        };
+
+        /// <summary>
+        /// Displays all properties of a model as a navigational menu selection with some additional filtering logic based on InstallerType.
+        /// </summary>
+        /// <typeparam name="T">Model type.</typeparam>
+        /// <param name="model">Instance of object model.</param>
+        /// <param name="exitMenuWord">Exit keyword to be shown to the user to exit the navigational menu.</param>
+        public static void DisplayPropertiesAsMenuSelection<T>(T model, string exitMenuWord)
```
palenshus, line 46: This name implies that it merely displays the properties, but it also allows the user to change the prop. I think something like PromptPropertiesWithMenu or something like that would be more apt

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+            selectionList.Add(Resources.DisplayPreview_MenuItem);
+            selectionList.Add(Resources.None_MenuItem);
+            return selectionList;
+        }
+
+        private static void ApplyChangesToIndividualInstallers(Installer installerCopy, List<Installer> installers)
+        {
+            var modifiedFields = installerCopy.GetType().GetProperties()
+                .Select(prop => prop)
+                .Where(pi => pi.GetValue(installerCopy) != null && pi.Name != nameof(Installer.Architecture));
```
palenshus, line 358: Comment on why arch is excluded

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+            string[] selection = selectedItem.Split('|');
+            var matchedInstallers = installerManifest.Installers.Where(
+                item => item.Architecture == (InstallerArchitecture)Enum.Parse(typeof(InstallerArchitecture), selection[0]) &&
+                item.InstallerType == (InstallerType)Enum.Parse(typeof(InstallerType), selection[1], true) &&
+                item.InstallerUrl == selection[2]);
+
+            Installer selectedInstaller;
+            if (matchedInstallers.Count() > 1)
+            {
+                Scope scope;
```
palenshus, line 299: can just declare inline on line 306

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+                    }
+                    else
+                    {
+                        installerLocale = selection[i];
+                    }
+                }
+
+                selectedInstaller = matchedInstallers.Single(
+                    item => nullableScope != null && item.Scope == nullableScope &&
+                    !string.IsNullOrEmpty(installerLocale) && item.InstallerLocale == installerLocale);
```
palenshus, line 318: won't this fail if scope has a value but locale doesn't?

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -0,0 +1,270 @@
...
+            }
+
+            var value = Prompt.Select(message, enumList, defaultValue: defaultValue);
+            if (value != Resources.None_MenuItem)
+            {
+                property.SetValue(model, Enum.Parse(enumType, value));
+            }
+        }
+
+        private static void PromptForPackageDependencies(List<PackageDependencies> packageDependencies)
```
palenshus, line 222: >PromptForPackageDependencies [](http://example.com/codeflow?start=28&length=28)

Can this be made more generic? I'm thinking a method that takes in a List<T> and a resource message where T : null()?

Not totally sure that'd work, but it'd be cool to have this method already be flexible to use with other future types

src/WingetCreateCore/Common/Extensions.cs
```diff
@@ -53,5 +57,51 @@ public static bool EqualsIC(this string a, string b)
...
+                       o.GetType().IsGenericType &&
+                       o.GetType().GetGenericTypeDefinition().IsAssignableFrom(typeof(Dictionary<,>));
+        }
+
+        /// <summary>
+        /// Determines if the type is a List type.
+        /// </summary>
+        /// <param name="type">Type to be evaluated.</param>
+        /// <returns>Boolean value indicating whether the type is a List.</returns>
+        public static bool IsList(this Type type)
```
palenshus, line 78: >IsList [](http://example.com/codeflow?start=27&length=6)

Doesn't this tell you if it's IsEnumerable instead?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,130 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+                    PromptHelper.DisplayPropertiesAsMenuSelection(selectedInstaller, Resources.None_MenuItem);
+                }
+            }
+        }
+
+        private static Installer MatchMenuSelectionToInstaller(string selectedItem, InstallerManifest installerManifest)
+        {
+            string[] selection = selectedItem.Split('|');
+            var matchedInstallers = installerManifest.Installers.Where(
+                item => item.Architecture == (InstallerArchitecture)Enum.Parse(typeof(InstallerArchitecture), selection[0]) &&
```
palenshus, line 292: >Enum.Parse( [](http://example.com/codeflow?start=68&length=11)

Can use Enum.Parse<InstallerArchitecture(selection[0])

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -0,0 +1,270 @@
...
+                {
+                    var value = Prompt.List<string>(message, minimum: minimum, validators: new[] { FieldValidation.ValidateProperty(validationModel, validationName, defaultValue) });
+                    if (!value.Any())
+                    {
+                        value = null;
+                    }
+
+                    property.SetValue(model, value);
+                }
+                else if (elementType == typeof(int))
```
palenshus, line 151: Can you make this more generic? What if the property was a double, or long, etc?

src/WingetCreateCore/Common/Extensions.cs
```diff
@@ -53,5 +57,51 @@ public static bool EqualsIC(this string a, string b)
...
+        }
+
+        /// <summary>
+        /// Determines if the properties of an object are all equal to null excluding properties with dictionary type if needed.
+        /// </summary>
+        /// <param name="o">Object to be evaluated.</param>
+        /// <returns>Boolean value indicating whether the object is empty.</returns>
+        public static bool IsEmptyObject(this object o)
+        {
+            return !o.GetType().GetProperties().Select(pi => pi.GetValue(o)).Where(value => !value.IsDictionary() && value != null).Any();
```
palenshus, line 104: Why is dictionary special here?

palenshus:
> // Copyright (c) Microsoft. All rights reserved.

Looks like only white-space changes in here?

---
Refers to: src/WingetCreateCore/Common/GitHub.cs:1 in e8bd161. [](commit_id = e8bd161042d858788567f346b3d2f3a526e74d1f, deletion_comment = True)

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -0,0 +1,270 @@
...
+                    }
+                    else
+                    {
+                        property.SetValue(model, null);
+                    }
+                }
+            }
+            else
+            {
+                if (property.PropertyType == typeof(InstallerSwitches))
```
palenshus, line 193: Asking same question, can this be made generic?

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: I agree, changed to suggestion

ryfu-msft: Didn't know you could do this, changed to suggestion

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Changed ordering so URL shows up last

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: Added an inclusion and exclusion list to handle this special cases. I haven't messed with Platform as I need further clarification. Do you mean that we should be excluding Platform for everything but MSIX and setting Platform: Desktop by default for all other installers except MSIX? Or is Platform:Desktop the default for all other installers regardless of whether its present in the manifest.

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: I like that better, changed to suggestion

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: added comment explanation

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: changed to suggestion

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: declare inline

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: I changed the logic for this method around, this shouldn't be an issue anymore

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: This was a lot harder to do, but I think I got it working, all the prompting methods should be generic now.

src/WingetCreateCore/Common/Extensions.cs
ryfu-msft: Yes, you're right, I meant to check for IsEnumerable and not list. Changed extension name

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: Made generic

src/WingetCreateCore/Common/Extensions.cs
ryfu-msft: Because of the schema, all object models include an "AdditionalProperties" field which is a dictionary. I don't foresee us needing to allow users to ever utilize that, hence why I exclude dictionaries by default.

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: Made generic

ryfu-msft:
> // Copyright (c) Microsoft. All rights reserved.

reverted

---
In reply to: [899274173](https://github.com/microsoft/winget-create/pull/133#issuecomment-899274173) [](http://example.com/codeflow?ancestors=899274173,899274173)

---
Refers to: src/WingetCreateCore/Common/GitHub.cs:1 in e8bd161. [](commit_id = e8bd161042d858788567f346b3d2f3a526e74d1f, deletion_comment = True)

src/WingetCreateCLI/PromptHelper.cs
palenshus: I think we need clarification from @denelon on Platform. I'm not sure when it would be anything but Desktop. Also not sure if it should be editable. For MSIX we get it from the appxmanifest, and I'm not sure if it applies to any other installer types.

PackageFamilyName shouldn't be editable, we get it from the appxmanifest.

ProductCode should be editable for "exe" installer type too, due to embedded MSIs.

src/WingetCreateCLI/PromptHelper.cs
palenshus: Wow, above and beyond! :)

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -100,10 +100,7 @@ public override async Task<bool> Execute()
 
                 if (!this.InstallerUrls.Any())
                 {
-                    this.InstallerUrls = PromptProperty(
-                        new Installer(),
-                        this.InstallerUrls,
-                        nameof(Installer.InstallerUrl));
+                    PromptHelper.PromptProperty(this, nameof(this.InstallerUrls), this.InstallerUrls, minimum: 1, validationModel: new Installer(), validationName: nameof(Installer.InstallerUrl));
```
palenshus, line 103: >PromptProperty [](http://example.com/codeflow?start=33&length=14)

Doesn't this still set the value too?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,136 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+                    Console.WriteLine();
+                    Logger.InfoLocalized(nameof(Resources.DisplayPreviewOfSelectedInstaller_Message));
+                    var serializer = Serialization.CreateSerializer();
+                    string installerString = serializer.Serialize(installerManifest);
+                    Console.WriteLine(installerString);
+                    Console.WriteLine();
+                }
+                else
+                {
+                    Installer selectedInstaller = MatchMenuSelectionToInstaller(selectedItem, installerManifest);
```
palenshus, line 282: Sorry, I didn't think of this before, but what if instead of trying to reverse-engineer the string to get the Installer, you just create a temporary Dictionary of tuplestring -> Installer, and then after the user selects one, you can look it up in the dictionary to get the Installer? Seems like that'd be less error-prone and easier to maintain

palenshus (dismissed):
:clock1:

src/WingetCreateCLI/PromptHelper.cs
denelon: Platform is in place for a future where we might be installing on other devices like HoloLens, XBox or the next computing platform. I think it would be pretty awesome to be able to `winget install <package>` on a quantum computer 😊.

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Yes you're right, I wanted the name to be as concise as possible, but I think setting the value is an important behavior to note. Reverted to "PromptPropertyAndSetValue"

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Good idea, changed to suggestion

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: I removed PackageFamilyName from msix since it shouldn't be editable and we can get that from appxmanifest.

Since platform isn't fully ready yet, I made it hidden for now for all installers except MSIX.

I also allowed ProductCode to be editable if installertype is an EXE.

palenshus: Cool. So @ryfu-msft I think you made the right choice to just make Platform non-editable for now. Someone can set it in their appx manifest, but we shouldn't allow it to be editable

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -237,10 +239,89 @@ private static void PromptRequiredProperties<T>(T manifest, VersionManifest vers
...
+                        installer.Scope?.ToEnumAttributeValue(),
+                        installer.InstallerLocale,
+                        installer.InstallerUrl,
+                    }.Where(s => !string.IsNullOrEmpty(s)));
+
+                installerSelectionMap.Add(installerTuple, installer);
+            }
+
+            List<string> selectionList = new List<string>() { Resources.AllInstallers_MenuItem };
+            selectionList.AddRange(new List<string>(installerSelectionMap.Keys));
```
palenshus, line 307: nit: You don't need to create a new list here in the AddRange

src/WingetCreateCLI/Commands/NewCommand.cs
palenshus: Nice, looks a lot better!

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: removed new list from addrange

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: Got it, so it looks like the short answer is don't make platform editable at all for anything right now. Removed platform from editable list.

palenshus (approved):
:shipit:

PR #138, opened 2021-08-17
Add Installer Exceptions for handling matching exceptions from core library

### LINKED ISSUE #101: Add localization to WingetCreateCore to provide better error messages from library functions

Since we don't have localization in the core library, we have to bend over backwards to provide helpful errors to the user from errors in the library function. The library functions need to somehow convey their error to the calling function so it can display localized errors instead of just logging it itself, or throwing a localized exception.

Add localization to the core library, and simplify error handling to log some error messages directly from the library.

ryfu-msft:
Resolves #101 

Instead of adding localization to the core library as a means for conveying an error to the calling function, I believe our issues can be fixed by throwing custom exceptions and catching those thrown exceptions in the calling function. As of right now, there hasn't been a need to add any logging functionality to the core library as all of the logging takes place in the CLI and we should try to maintain that design choice until we absolutely need to make that change.

Changes:
- Added custom InstallerExceptions for MultipleMatchedInstallerException and UnmatchedInstallerException which both take in an Installer parameter to better pass information about the installer causing the matching error.
- Handle these unique exceptions in the calling function in UpdateCommand.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/138)


src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -199,43 +200,44 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
+            {
+                if (!PackageParser.UpdateInstallerNodesAsync(
+                    installerManifest,
+                    this.InstallerUrls,
+                    packageFiles,
+                    out detectedArchOfInstallers))
                 {
-                    Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
+                    DisplayMismatchedArchitectures(detectedArchOfInstallers);
+                    Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error));
```
jamespik, line 214: This still looks like some errrors are return via exceptions, and others via exit code.  Why do we have split error handling?

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -288,6 +288,12 @@ protected static async Task<string> DownloadPackageFile(string installerUrl)
                     return null;
                 }
 
+                if (e is InvalidDataException)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.DownloadFileExceedsMaxSize_Error), e.Message);
```
jamespik, line 293: Instead of just putting this in the message i rather like making these strange typed values on the exception, like what you did for the other exception. In that manner, we can still jhave a meaningful message on the exception for debugging, but the data the caller needs to derive a good localized error message can still be handed out?

jamespik (changes requested):

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -275,7 +267,14 @@ public static string GetFileHash(string path)
                 matchingExistingInstaller.Platform = newInstaller.Platform;
             }
 
-            installerMismatch = unmatchedInstallers.Any() || multipleMatchedInstallers.Any();
+            if (installerMismatch = unmatchedInstallers.Any() || multipleMatchedInstallers.Any())
+            {
+                List<Exception> exceptionList = new List<Exception>();
+                unmatchedInstallers.ForEach(i => exceptionList.Add(new InstallerExceptions.UnmatchedInstallerException(i)));
+                multipleMatchedInstallers.ForEach(i => exceptionList.Add(new InstallerExceptions.MultipleMatchedInstallerException(i)));
+                throw new AggregateException(exceptionList);
```
palenshus, line 275: Instead of an aggregate exception, what about an InstallerMatchException or something like that, with two properties... List<Installer> MultipleMatchedInstallers and List<Installer> UnmatchedInstallers?

That seems more concise to me, but tell me if you disagree

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I created a new exception for DownloadSizeExceeded similar to the other exceptions I created so that the data value isn't hidden in the exception message.

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: totally agree. I created an InstallerMatchException that does just that. should be cleaner now, thanks for the suggestion.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: I believe in this case that our errors should be handled with exceptions. That way we can obtain better information about the specific packages that failed parsing. I modified the method so that errors are handled consistently.

src/WingetCreateCore/Common/Exceptions.cs
```diff
@@ -0,0 +1,82 @@
...
+namespace Microsoft.WingetCreateCore.Common
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.WingetCreateCore.Models.Installer;
+
+    /// <summary>
+    /// Exception class for handling installer matching errors.
+    /// </summary>
+    public class Exceptions
```
palenshus, line 13: Weird to have this nested like this. Instead why not just extend the namespace to be Microsoft.WingetCreateCore.Common.Exception instead?

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -199,42 +200,43 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
-                        Logger.ErrorLocalized(nameof(Resources.MultipleMatchingInstallerNodes_Error));
-                        Console.WriteLine();
-                        multipleMatchedInstallers.ForEach(i =>
-                        {
-                            Logger.ErrorLocalized(nameof(Resources.InstallerDetectedFromUrl_Message), i.Architecture, i.InstallerType, i.InstallerUrl);
-                            Console.WriteLine();
-                        });
-                    }
+                if (multipleMatchedInstallers.Any())
+                {
```
palenshus, line 233: This is redundant, the ForEach won't do anything if there's !Any

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -199,42 +200,43 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
+                    out detectedArchOfInstallers);
+            }
+            catch (InvalidOperationException)
+            {
+                Logger.ErrorLocalized(nameof(Resources.InstallerCountMustMatch_Error));
+                return null;
+            }
+            catch (Exceptions.ParsePackageException parsePackageException)
+            {
+                var parseFailedInstallerUrls = parsePackageException.ParseFailedInstallerUrls;
```
palenshus, line 220: [nit] IMO, assigning this to a new variable here isn't worth it, and below

src/WingetCreateCore/Common/Exceptions.cs
ryfu-msft: Extended namespace and created an Exceptions folder containing the exception classes.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Removed Any check

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: remove assignment to new variable

src/WingetCreateCLI/Commands/BaseCommand.cs
jamespik: Lol.. "Strong" not strange   :)

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -282,9 +283,15 @@ protected static async Task<string> DownloadPackageFile(string installerUrl)
...
+                if (e is HttpRequestException httpRequestException)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.HttpResponseUnsuccessful_Error), exception.StatusCode);
+                    Logger.ErrorLocalized(nameof(Resources.HttpResponseUnsuccessful_Error), httpRequestException.StatusCode);
+                    return null;
+                }
+
+                if (e is DownloadSizeExceededException downloadSizeExceededException)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.DownloadFileExceedsMaxSize_Error), $"{downloadSizeExceededException.MaxDownloadSize / 1024 / 1024}");
```
jamespik, line 294: Consider indicating the expected Units on the variable name. Ie MaxDownloadSizeInWhatever

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -199,44 +200,33 @@ public async Task<Manifests> DeserializeExistingManifestsAndUpdate(List<string>
...
+            }
+            catch (ParsePackageException parsePackageException)
+            {
+                parsePackageException.ParseFailedInstallerUrls.ForEach(i => Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error), i));
+                return null;
+            }
+            catch (InstallerMatchException installerMatchException)
+            {
+                Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
+                Logger.ErrorLocalized(nameof(Resources.InstallersFailedToMatch_Message));
```
jamespik, line 226: Are these ever used independently? If not, consider merging the resources as it adds more context for better localization.

src/WingetCreateCLI/Commands/BaseCommand.cs
jamespik: Max download size is nullable. C# i believe that will force the entire arithmetic to null.  should the download size not be optional?

jamespik (changes requested):

src/WingetCreateCLI/Commands/BaseCommand.cs
palenshus: This exception will only ever be thrown if MaxDownloadSize is not null (see PackageParser.cs)

palenshus (approved):
:shipit:

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Changed variable name to include expected units. I also changed the type of MaxDownloadSize to not be nullable so that the arithmetic doesn't compute to null. I believe that this is okay since DownloadSizeExceededException will only be called if a maxdownloadsize is properly defined.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: They were before, but not anymore. I merged the resources into a single resource string so that context is more clear.

jamespik (approved):

PR #162, opened 2021-09-09
Add additional help messages and make error warnings non-blocking

ryfu-msft:
This pull request adds additional help messages to better instruct the user what to do when an installer mismatch is encountered when using the update command autonomously.

Also makes these scenarios non-blocking:
- Duplicate package id detected (Now asks the user to confirm if they would like to continue)
- Invalid repository path now shows an error message but is no longer blocking (fixes a bug where the settings command could not be executed because of this error)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/162)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -263,7 +263,6 @@ namespace Microsoft.WingetCreateCLI.Commands
         protected static async Task<string> DownloadPackageFile(string installerUrl)
         {
             Logger.InfoLocalized(nameof(Resources.DownloadInstaller_Message), installerUrl);
-            Console.WriteLine();
 
             try
             {
@@ -431,17 +430,17 @@ namespace Microsoft.WingetCreateCLI.Commands
                 if (e is AuthorizationException)
                 {
                     Logger.ErrorLocalized(nameof(Resources.InvalidGitHubToken_Message));
+                    return false;
                 }
                 else if (e is RateLimitExceededException)
                 {
                     Logger.ErrorLocalized(nameof(Resources.RateLimitExceeded_Message));
+                    return false;
                 }
                 else if (e is NotFoundException)
                 {
                     Logger.ErrorLocalized(nameof(Resources.RepositoryNotFound_Error), this.WingetRepoOwner, this.WingetRepo);
                 }
-
-                return false;
             }
 
             this.GitHubClient = client;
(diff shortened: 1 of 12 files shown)


src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -431,17 +430,17 @@ protected async Task<bool> CheckGitHubTokenAndSetClient()
...
                 {
                     Logger.ErrorLocalized(nameof(Resources.RateLimitExceeded_Message));
+                    return false;
                 }
                 else if (e is NotFoundException)
                 {
                     Logger.ErrorLocalized(nameof(Resources.RepositoryNotFound_Error), this.WingetRepoOwner, this.WingetRepo);
                 }
-
-                return false;
```
palenshus, line 444: Instead of changing this to succeed on rep not found, why don't we just not call LoadGitHubClient for SettingsCommand (and CacheCommand)?

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -141,8 +141,10 @@ public override async Task<bool> Execute()
                         if (!await this.PromptPackageIdentifierAndCheckDuplicates(manifests))
                         {
                             Console.WriteLine();
-                            Logger.ErrorLocalized(nameof(Resources.PackageIdAlreadyExists_Error));
-                            return false;
+                            if (!Prompt.Confirm(Resources.PackageIdAlreadyExists_Message))
```
palenshus, line 144: What's the rationale for making this non-blocking?

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: I think that makes more sense. I had to make that change in the program.cs since that is where it is being called. The github client is no longer loaded for the cache/settings command

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Kevin did not like that he could not create a manifest that already existed. I think this is the best compromise to notify the user that their package is a duplicate while still allowing them to use our tool to generate any manifest they want

palenshus: As discussed offline with @KevinLaMS, we shouldn't allow users to bypass this. For testing purposes, we can always manipulate the string to avoid a duplicate

ryfu-msft: Reverted changes

ryfu-msft: Reverted changes.

palenshus (approved):
:shipit:

PR #164, opened 2021-09-10
Fix pass by reference bug when applying changes to all installers

ryfu-msft:
This PR addresses a bug where List objects passed by reference were causing errors when serializing the manifest as seen in the image below (InstallerSuccessCodes produces weird characters):

MicrosoftTeams-image

The fix was to assign the object by value rather than by reference which requires creating new instances of the edited values

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/164)

### DIFF
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.WingetCreateCLI.Commands
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.ComponentModel.DataAnnotations;
     using System.IO;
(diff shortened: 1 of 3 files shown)


src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -321,7 +322,55 @@ private static void ApplyChangesToIndividualInstallers(Installer installerCopy,
...
+                    {
+                        prop.SetValue(installer, fieldValue);
+                    }
+                    else if (fieldValue is IList)
+                    {
+                        Type elementType = prop.PropertyType.GetGenericArguments()[0];
+                        Type listType = typeof(List<>).MakeGenericType(elementType);
+                        var list = Activator.CreateInstance(listType, fieldValue);
+                        prop.SetValue(installer, list);
+                    }else if (fieldValue is Dependencies dependencies)
```
palenshus, line 336: formatting

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -321,7 +322,55 @@ private static void ApplyChangesToIndividualInstallers(Installer installerCopy,
...
+                if (fieldValue is List<PackageDependencies> packageDependencies)
+                {
+                    List<PackageDependencies> packageDependenciesList = new List<PackageDependencies>();
+                    packageDependencies.ForEach(i => packageDependenciesList.Add(
+                        new PackageDependencies { PackageIdentifier = i.PackageIdentifier, MinimumVersion = i.MinimumVersion }));
+                    prop.SetValue(installer.Dependencies, packageDependenciesList);
+                }
+                else if (fieldValue is IList)
+                {
+                    Type elementType = prop.PropertyType.GetGenericArguments()[0];
```
palenshus, line 370: we have this code twice, it should be a method

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Fixed

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: No longer relevant with your new changes

palenshus (approved):
:shipit:

PR #165, opened 2021-09-12
Handle installers blocked by defender virus scan

ryfu-msft:
This PR adds the following changes:

- Additional documentation in the security.md file regarding Virus and Threat Protection
- Catches IOException which is thrown when a downloaded installer is blocked by defender scan.
- Reformatting Exception handling

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/165)

### DIFF
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -116,9 +116,17 @@ namespace Microsoft.WingetCreateCLI.Commands
                     PackageParser.ParsePackages(packageFiles, this.InstallerUrls, manifests, out List<PackageParser.DetectedArch> detectedArchs);
                     DisplayMismatchedArchitectures(detectedArchs);
                 }
-                catch (ParsePackageException exception)
+                catch (Exception e)
                 {
-                    exception.ParseFailedInstallerUrls.ForEach(i => Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error), i));
+                    if (e is ParsePackageException exception)
+                    {
+                        exception.ParseFailedInstallerUrls.ForEach(i => Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error), i));
+                    }
+                    else if (e is IOException)
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.DefenderVirus_ErrorMessage));
+                    }
+
                     return false;
                 }
 
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -117,25 +117,14 @@ namespace Microsoft.WingetCreateCLI.Commands
                     PackageParser.ParsePackages(packageFiles, this.InstallerUrls, manifests, out List<PackageParser.DetectedArch> detectedArchs);
                     DisplayMismatchedArchitectures(detectedArchs);
                 }
-                catch (Exception e)
+                catch (IOException iOException) when (iOException.HResult == -2147024671)
                 {
-                    if (e is ParsePackageException parsePackageException)
-                    {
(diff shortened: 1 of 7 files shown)


denelon (approved):

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -116,9 +116,17 @@ public override async Task<bool> Execute()
...
                 }
-                catch (ParsePackageException exception)
+                catch (Exception e)
                 {
-                    exception.ParseFailedInstallerUrls.ForEach(i => Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error), i));
+                    if (e is ParsePackageException exception)
+                    {
+                        exception.ParseFailedInstallerUrls.ForEach(i => Logger.ErrorLocalized(nameof(Resources.PackageParsing_Error), i));
+                    }
+                    else if (e is IOException)
```
palenshus, line 125: There could be other reasons we hit an IOException, it's a pretty broad exception: [https://docs.microsoft.com/en-us/dotnet/standard/io/handling-io-errors](https://docs.microsoft.com/en-us/dotnet/standard/io/handling-io-errors)

Is there a specific property (HResult, message, innerexception, etc) on the exception that you can use to narrow down to this scenario?

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Added an additional if statement that checks the HRESULT of the exception to determine if its a virus exception.

palenshus: you can use an exception filter to avoid the else clause: [https://docs.microsoft.com/en-us/dotnet/standard/exceptions/using-user-filtered-exception-handlers](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/using-user-filtered-exception-handlers)

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -247,24 +247,27 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
...
-                Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
-                installerMatchException.MultipleMatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.UnmatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
-                installerMatchException.UnmatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.MultipleMatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
-                Console.WriteLine();
-                Logger.WarnLocalized(nameof(Resources.ResolveMatchingConflicts_Message));
+                if (e is InvalidOperationException)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.InstallerCountMustMatch_Error));
+                }
+                else if (e is IOException)
```
palenshus, line 256: Needs filter here too

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Oops, I guess I missed that. Added filter

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft: Added exception filter

palenshus (approved):
:shipit:

PR #168, opened 2021-09-15
Handle download installer connection failure

ryfu-msft:
Changes:

- Adds a special error message if the httprequestexception is detected to be a connection attempt failure (repro: wingetcreate new http://foo)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/168)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -284,7 +284,15 @@ namespace Microsoft.WingetCreateCLI.Commands
 
                 if (e is HttpRequestException httpRequestException)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.HttpResponseUnsuccessful_Error), httpRequestException.StatusCode);
+                    if (httpRequestException.HResult == -2147467259)
+                    {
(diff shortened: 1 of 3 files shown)


src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -284,7 +284,15 @@ protected static async Task<string> DownloadPackageFile(string installerUrl)
 
                 if (e is HttpRequestException httpRequestException)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.HttpResponseUnsuccessful_Error), httpRequestException.StatusCode);
+                    if (httpRequestException.HResult == -2147467259)
```
palenshus, line 287: This is just E_FAIL, which is a really generic error code. I'd instead check if StatusCode != null then display the original error message, otherwise just display ex.Message

ryfu-msft: Changed to suggestion. Added check for StatusCode != null instead.

palenshus (approved):
:shipit:

PR #171, opened 2021-09-16
Add modify feature to interactive update

### LINKED ISSUE #111: Add functionality to modify an existing manifest
As of right now, we only have the new command to generate new manifests and the update command to update installer urls, version, and hash. There is a need for a `modify` command that pulls down an existing manifest and allows a user to make changes to any field they want to edit. 

This would address issues like #75 by allowing users to replace installers and their related fields. We currently don't support replacing existing installers with new ones in order to preserve updating multiple installers and avoid accidentally replacing the wrong installer.

The `modify` command flow would look something like this:

`wingetcreate modify <Package Identifier> -v <Version>`

- Pulls down the manifests for the specified package and the specified version (if no version is specified, latest manifest is pulled down)
- Prompts would show for each field, with some form of selection logic to allow users to edit by manifest, installer node, field, etc. 
- Saves newly modified manifests locally and also prompts the user if they would like to submit to the Winget-Pkgs repository.

ryfu-msft:
This PR resolves #111 

Changes:
- Added navigational menu for interactive update scenario. You can now navigate and select the manifest you want to edit and modify the fields as needed.
- Moved shared methods from NewCommand.cs to PromptHelper.cs 
- Changed the "NONE" exit keyword to "SAVE AND EXIT"

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/171)

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -371,6 +371,55 @@ namespace Microsoft.WingetCreateCLI.Commands
             return newHashes.Except(oldHashes).Any();
         }
 
+        private static void DisplayManifestsAsMenuSelection(Manifests manifests)
+        {
+            Console.Clear();
+            string versionFileName = Manifests.GetFileName(manifests.VersionManifest);
+            string installerFileName = Manifests.GetFileName(manifests.InstallerManifest);
+            string versionManifestMenuItem = $"{manifests.VersionManifest.ManifestType.ToUpper()}: " + versionFileName;
+            string installerManifestMenuItem = $"{manifests.InstallerManifest.ManifestType.ToUpper()}: " + installerFileName;
+
+            while (true)
+            {
+                // Need to update locale manifest file name each time as PackageLocale can change
+                string defaultLocaleMenuItem = $"{manifests.DefaultLocaleManifest.ManifestType.ToUpper()}: " + Manifests.GetFileName(manifests.DefaultLocaleManifest);
+                List<string> selectionList = new List<string> { versionManifestMenuItem, installerManifestMenuItem, defaultLocaleMenuItem };
+                Dictionary<string, LocaleManifest> localeManifestMap = new Dictionary<string, LocaleManifest>();
+                foreach (LocaleManifest localeManifest in manifests.LocaleManifests)
+                {
+                    string localeManifestFileName = $"{localeManifest.ManifestType.ToUpper()}: " + Manifests.GetFileName(localeManifest);
+                    localeManifestMap.Add(localeManifestFileName, localeManifest);
+                    selectionList.Add(localeManifestFileName);
+                }
+
+                selectionList.Add(Resources.Done_MenuItem);
+                var selectedItem = Prompt.Select(Resources.SelectManifestToEdit_Message, selectionList);
+
+                if (selectedItem == versionManifestMenuItem)
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.VersionManifest, Resources.SaveAndExit_MenuItem, versionFileName);
+                }
+                else if (selectedItem == installerManifestMenuItem)
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.InstallerManifest, Resources.SaveAndExit_MenuItem, installerFileName);
+                }
+                else if (selectedItem == defaultLocaleMenuItem)
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.DefaultLocaleManifest, Resources.SaveAndExit_MenuItem, Manifests.GetFileName(manifests.DefaultLocaleManifest));
+
+                }
+                else if (selectedItem == Resources.Done_MenuItem)
+                {
+                    break;
+                }
+                else
+                {
+                    var selectedLocaleManifest = localeManifestMap[selectedItem];
+                    PromptHelper.PromptPropertiesWithMenu(selectedLocaleManifest, Resources.SaveAndExit_MenuItem, Manifests.GetFileName(selectedLocaleManifest));
+                }
+            }
+        }
+
         /// <summary>
         /// Update flow for interactively updating the manifest.
         /// </summary>s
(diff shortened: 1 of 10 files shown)


src/WingetCreateCore/Common/Serialization.cs
```diff
@@ -137,6 +137,18 @@ public static string ConvertYamlToJson(string yaml)
             return JsonConvert.SerializeObject(yamlObject);
         }
 
+        /// <summary>
+        /// Serializes the provided object and returns the serialized string.
+        /// </summary>
+        /// <typeparam name="T">Object type.</typeparam>
+        /// <param name="value">Object to be serialized.</param>
+        /// <returns>Serialized string.</returns>
+        public static string Serialize<T>(T value)
```
palenshus, line 146: No need for this to be generic, it can just take an object

src/WingetCreateCore/Models/Manifests.cs
```diff
@@ -40,6 +40,40 @@ public class Manifests
...
+            else if (manifest is LocaleManifest localeManifest)
+            {
+                fileName = $"{localeManifest.PackageIdentifier}.locale.{localeManifest.PackageLocale}.yaml";
+            }
+            else if (manifest is SingletonManifest singletonManifest)
+            {
+                fileName = $"{singletonManifest.PackageIdentifier}.yaml";
+            }
+
+            return fileName;
```
palenshus, line 74: You can use a switch expression here:

            return manifest switch
            {
                InstallerManifest installerManifest => $"{installerManifest.PackageIdentifier}.installer.yaml",
                VersionManifest versionManifest => $"{versionManifest.PackageIdentifier}.yaml",
                DefaultLocaleManifest defaultLocaleManifest => $"{defaultLocaleManifest.PackageIdentifier}.locale.{defaultLocaleManifest.PackageLocale}.yaml",
                LocaleManifest localeManifest => $"{localeManifest.PackageIdentifier}.locale.{localeManifest.PackageLocale}.yaml",
                SingletonManifest singletonManifest => $"{singletonManifest.PackageIdentifier}.yaml",
                _ => throw new ArgumentException(nameof(manifest)),
            };

src/WingetCreateCore/Models/Manifests.cs
```diff
@@ -40,6 +40,40 @@ public class Manifests
...
+        }
+
         public List<Installer.Installer> CloneInstallers()
```
palenshus, line 77: public methods should be documented

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -370,6 +370,55 @@ private static bool VerifyUpdatedInstallerHash(Manifests oldManifest, InstallerM
...
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.VersionManifest, Resources.SaveAndExit_MenuItem, versionFileName);
+                }
+                else if (selectedItem == installerManifestMenuItem)
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.InstallerManifest, Resources.SaveAndExit_MenuItem, installerFileName);
+                }
+                else if (selectedItem == defaultLocaleMenuItem)
+                {
+                    PromptHelper.PromptPropertiesWithMenu(manifests.DefaultLocaleManifest, Resources.SaveAndExit_MenuItem, Manifests.GetFileName(manifests.DefaultLocaleManifest));
```
palenshus, line 407: Extra newline

palenshus:
>                     new Installer.Installer {

Can you not just use the CloneInstaller method for this?

---
Refers to: src/WingetCreateCore/Models/Manifests.cs:83 in a0a6e64. [](commit_id = a0a6e643bb85b28cb79480e1e72dad19a8f8f1dc, deletion_comment = False)

src/WingetCreateCLI/PromptHelper.cs
```diff
@@ -269,6 +309,119 @@ public static void PromptList<T>(string message, object model, string memberName
             }
         }
 
+        private static List<string> FilterRestrictedFields<T>(T model, PropertyInfo[] properties)
```
palenshus, line 312: This function is complex enough that it needs a documentation header

palenshus:
What's the goal of this feature? It appears it allows a user to update the installers one by one interactively, what fields does it update? I just tried it out and it appears it just updates the sha256 and url, is that correct?

palenshus (changes requested):
:clock1:

src/WingetCreateCore/Common/Serialization.cs
ryfu-msft: Changed parameter to take in an object

src/WingetCreateCore/Models/Manifests.cs
ryfu-msft: neat, changed to suggested. Thanks for the help :)

ryfu-msft:
>                     new Installer.Installer {

I honestly didn't even know that already existed. Changed to use existing CloneInstaller method

---
In reply to: [922043104](https://github.com/microsoft/winget-create/pull/171#issuecomment-922043104) [](http://example.com/codeflow?ancestors=922043104,922043104)

---
Refers to: src/WingetCreateCore/Models/Manifests.cs:83 in a0a6e64. [](commit_id = a0a6e643bb85b28cb79480e1e72dad19a8f8f1dc, deletion_comment = False)

src/WingetCreateCore/Models/Manifests.cs
ryfu-msft: Added documentation for public method

src/WingetCreateCLI/PromptHelper.cs
ryfu-msft: Added documentation headers for these 3 functions

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: removed newline

src/WingetCreateCore/Models/Manifests.cs
palenshus: Nice. Can you change basecommand.cs to use this method in the two places where we write files out too?

src/WingetCreateCLI/PromptHelper.cs
palenshus: Thanks!

src/WingetCreateCore/Models/Manifests.cs
ryfu-msft: Changed basecommand to use this method

ryfu-msft:
This feature is supposed to give users a way to resolve any matching conflicts when updating autonomously. 

Initially, I did not overwrite the architecture, but after further thought, it would make more sense to have the architecture accurately reflect installer that the user intends to update with. I have now changed the behavior of this in the ParsePackageAndUpdateInstallerNode method so that if we detect a new architecture, we update, otherwise, we default to the existing architecture. For appxbundles, I did not want have another matching experience based on the architectures of the installers detected in the bundle. I simply keep the existing architecture and update the other metadata. If there is an issue, the user can change the metadata in the navigational menu.

---
In reply to: [922050042](https://github.com/microsoft/winget-create/pull/171#issuecomment-922050042) [](http://example.com/codeflow?ancestors=922050042,922050042)

palenshus (approved):
:shipit:

PR #174, opened 2021-09-23
Validate manifest in interactive menu

### LINKED ISSUE #172: Interactive update menu does not validate manifest
When using the interactive update command to make changes on an existing manifest, a user has no way of knowing if the changes they made are valid.

Example:
- When changing the defaultLocale in the VersionManifest to a locale that doesn't have an associated localeManifest, Validation takes place after the manifest is generated and the user has no way of fixing that.

ryfu-msft:
Resolves #172 

Changes:
- Adds an extra validate manifest step prior to showing the menu selection of manifests. This informs the user if there's anything invalid and needs to be fixed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/174)

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -392,8 +392,8 @@ namespace Microsoft.WingetCreateCLI.Commands
                 }
 
                 selectionList.Add(Resources.Done_MenuItem);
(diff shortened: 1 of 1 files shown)


palenshus (approved):
:shipit:

PR #201, opened 2021-11-04
Handle taskcanceled exception with error message

### LINKED ISSUE #197: Unclear exception after timeout while downloading an installer

The error now just says "Couldn't download an installer" and spits out stack trace.

I think it would be better to explicitly say that timout happened, rather than make user try to read the trace, or add an option to extend the timeout duration.

Now it looks like this:
image

_No response_

ryfu-msft:
Fixes #197 

Changes
- Handles the exception that's thrown when the download request fails due to timeout.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/201)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -305,6 +305,12 @@ namespace Microsoft.WingetCreateCLI.Commands
                     return null;
                 }
 
+                if (e is TaskCanceledException)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.DownloadConnectionTimeout_Error));
+                    return null;
+                }
+
                 throw;
             }
         }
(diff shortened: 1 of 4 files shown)


src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -305,6 +305,12 @@ protected static async Task<string> DownloadPackageFile(string installerUrl)
                     return null;
                 }
 
+                if (e is TaskCanceledException)
```
palenshus, line 308: For clarity can you make these all if-else blocks?

palenshus (approved):

src/WingetCreateCLI/Commands/BaseCommand.cs
ryfu-msft: Added else-if blocks for clarify

palenshus (approved):

mohmadalabdalui (commented):
``

PR #206, opened 2021-11-24
Add architecture override option to Update command

### LINKED ISSUE #198: wingetcreate looks for installer architecture instead of application architecture

To replace my team's current custom tooling for deploying to `winget`, I am trying to adopt `wingetcreate` for automatic updates of the `Microsoft.Git` manifest each time we release. The manifest currently has one installer and specifies an `x64` architecture. Note that this refers to the _application to be installed_ not the _installer itself_ (which is `x86`).

When we attempt to publish a new manifest, `wingetcreate` downloads the installer and detects that the _installer_ is `x86`. It then compares it to the `x64` specified in the manifest and fails. I've included more details below, but in short I think `wingetcreate` is matching on __installer__ architecture when it should be matching on __application__ architecture.

### Steps to reproduce

__Pre-Requisites__
1. Use a Windows 10 machine
0. Ensure `winget` and `wingetcreate` are installed and on the path
0. Generate a PAT that has `repo:public_repo` scope

__Steps to Reproduce__

Run the following command (replacing *** with the token you generated above):

`wingetcreate update Microsoft.Git -u https://github.com/ldennington/git/releases/download/v2.33.0.vfs.0.7/Git-2.33.0.vfs.0.7-64-bit.exe -v 2.33.0.0.7 -o manifests -t ***`

### Expected behavior

I expected the `Microsoft.Git` manifest to be updated successfully and saved to the `manifests` directory.

### Actual behavior

The following error is thrown:

```
Each new installer URL must have a single match to an existing installer based on installer type and architecture. The following installers failed to match an existing installer:
Multiple matches found for X86 Inno installer detected from the url: https://github.com/ldennington/git/releases/download/v2.33.0.vfs.0.7/Git-2.33.0.vfs.0.7-64-bit.exe
```
I tried to update the architecture to `x86` in the manifest, but that broke `winget` validation (which makes sense since `x64` applications can't run on `x86` machines). So I don't actually think this was the right fix. My understanding is that Inno Setup only ships an `x86` version, and it's only possible to create `x86` Inno installers using this tech, so I would expect this issue to impact a lot of products beyond just `Microsoft.Git`.

In light of this, I believe that `wingetcreate` should be looking at the __application__ architecture instead of the __installer__ architecture when attempting to find an installer match. 

### Environment

```shell
Windows Package Manager Version: v1.1.12701
Windows OS Build: 19043.1288
wingetcreate Version: 0.4.2.1
Package: Microsoft.Git
```

ryfu-msft:
Resolves #198 

This PR addresses issues that some users have been having with the update command detecting the wrong architecture from the installers they provide when trying to use the update command. This discrepancy is often seen in nullsoft installers, which are detected to be x86 based off their installer executables, even if they install a x64 application.

To resolve this, we added the ability for users to append the overriding architecture to the end of the installerURL they are updating with. The architecture must be separated by the '|' character with no spaces.

Example (overriding x64 installer with x86):
wingetcreate update <PackageIdentifier> -u https://foobar_x64_installer.exe|x86

Changes:
- Adds an example to the UpdateCommand on how to use the architecture override.
- Add functionality for parsing installerUrls for architecture overrides by splitting based on the '|' character.
- Matching new installers to existing installers now first checks if a match can be found by comparing the installerType and the new overriding architecture. If no match can be found, then the architecture detected from the url string or the binary are used instead. If both of these comparisons fail, a matching error is shown to the user.
- Refactor matching logic into separate method called FindInstallerMatch().

Testing:
Added 5 unit tests to verify the functionality of this new features:
- Verifies that invalid architecture specified for override shows error message
- Verifies that multiple architectures specified for override shows error message
- Verifies update command succeeds with an overriding architecture specified that matches an existing installer
- Verifies the update command succeeds if the overriding architecture fails to match, but the url and binary detection does find a match to an existing installer.
- Verifies that failing a match when updating with an architecture override shows a warning message to the user.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/206)

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -249,20 +249,21 @@ namespace Microsoft.WingetCreateCLI.Commands
                 return null;
             }
 
-            List<PackageParser.DetectedArch> detectedArchOfInstallers;
-            List<Installer> newInstallers = new List<Installer>();
+            InstallerUpdateHelper installerUpdateHelper =
+                new InstallerUpdateHelper
+                {
+                    PackageFiles = packageFiles,
+                    InstallerUrls = this.InstallerUrls,
+                    NewInstallers = new List<Installer>(),
+                    DetectedArchitectures = new List<PackageParser.DetectedArch>(),
+                    ArchitectureOverrideMap = installerArchOverrideMap,
+                };
 
             try
             {
-                PackageParser.UpdateInstallerNodesAsync(
-                    installerManifest,
-                    this.InstallerUrls,
-                    packageFiles,
-                    out detectedArchOfInstallers,
-                    out newInstallers,
-                    installerArchOverrideMap);
-
-                DisplayMismatchedArchitectures(detectedArchOfInstallers);
+                // List of detected architectures should get populated by update.
+                PackageParser.UpdateInstallerNodesAsync(installerManifest, installerUpdateHelper);
+                DisplayMismatchedArchitectures(installerUpdateHelper.DetectedArchitectures);
             }
             catch (InvalidOperationException)
             {
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -224,6 +224,12 @@ namespace Microsoft.WingetCreateCLI.Commands
             // Parse out architecture overrides from installer URLs and reassign.
             this.InstallerUrls = this.ParseInstallerUrlsForArchOverride(this.InstallerUrls.ToList(), out Dictionary<string, InstallerArchitecture> installerArchOverrideMap);
 
+            // If InstallerUrls is null, there was an issue when parsing for architecture override.
+            if (this.InstallerUrls == null)
+            {
+                return null;
+            }
+
             foreach (var key in installerArchOverrideMap.Keys)
             {
                 Logger.Warn($"Overriding {key} with architecture {installerArchOverrideMap[key]}");
@@ -447,32 +453,36 @@ namespace Microsoft.WingetCreateCLI.Commands
         {
             installerArchOverrideMap = new Dictionary<string, InstallerArchitecture>();
             List<string> parsedInstallerUrls = new List<string>();
-            foreach (string installerUrl in installerUrlsToBeParsed)
+            foreach (string item in installerUrlsToBeParsed)
             {
-                if (installerUrl.Contains('|'))
+                if (item.Contains('|'))
                 {
                     // '|' character indicates that an architecture override can be parsed from the installer.
-                    string[] installerUrlOverride = installerUrl.Split('|');
+                    string[] installerUrlOverride = item.Split('|');
 
                     if (installerUrlOverride.Length > 2)
                     {
                         Logger.ErrorLocalized(nameof(Resources.MultipleArchitectureOverride_Error));
+                        return null;
                     }
 
-                    InstallerArchitecture? overrideArch = installerUrlOverride[1].ToEnumOrDefault<InstallerArchitecture>();
+                    string installerUrl = installerUrlOverride[0];
+                    string overrideArchString = installerUrlOverride[1];
+                    InstallerArchitecture? overrideArch = overrideArchString.ToEnumOrDefault<InstallerArchitecture>();
                     if (overrideArch.HasValue)
                     {
-                        parsedInstallerUrls.Add(installerUrlOverride[0]);
-                        installerArchOverrideMap.Add(installerUrlOverride[0], overrideArch.Value);
+                        parsedInstallerUrls.Add(installerUrl);
+                        installerArchOverrideMap.Add(installerUrl, overrideArch.Value);
                     }
                     else
                     {
-                        Logger.ErrorLocalized(nameof(Resources.UnableToParseArchOverride_Error), installerUrlOverride[1]);
+                        Logger.ErrorLocalized(nameof(Resources.UnableToParseArchOverride_Error), overrideArchString);
+                        return null;
                     }
                 }
                 else
                 {
-                    parsedInstallerUrls.Add(installerUrl);
+                    parsedInstallerUrls.Add(item);
                 }
             }
 
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -41,6 +41,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             {
                 yield return new Example(Resources.Example_UpdateCommand_SearchAndUpdateVersionAndInstallerURL, new UpdateCommand { Id = "<PackageIdentifier>", InstallerUrls = new string[] { "<InstallerUrl1>", "<InstallerUrl2>" }, Version = "<Version>" });
                 yield return new Example(Resources.Example_UpdateCommand_SaveAndPublish, new UpdateCommand { Id = "<PackageIdentifier>", Version = "<Version>", OutputDir = "<OutputDirectory>", GitHubToken = "<GitHubPersonalAccessToken>" });
+                yield return new Example(Resources.Example_UpdateCommand_OverrideArchitecture, new UpdateCommand { Id = "<PackageIdentifier>", InstallerUrls = new string[] { "<InstallerUrl1>|<InstallerArchitecture>" }, Version = "<Version>" });
             }
         }
 
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -450,10 +450,10 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// <returns>List of <see cref="InstallerMetadata"/> helper objects used for updating the installers.</returns>
         private List<InstallerMetadata> ParseInstallerUrlsForArchOverride(List<string> installerUrlsToBeParsed)
         {
-            List<InstallerMetadata> installerUpdateList = new List<InstallerMetadata>();
+            List<InstallerMetadata> installerMetadataList = new List<InstallerMetadata>();
             foreach (string item in installerUrlsToBeParsed)
             {
-                InstallerMetadata installerUpdate = new InstallerMetadata();
+                InstallerMetadata installerMetadata = new InstallerMetadata();
 
                 if (item.Contains('|'))
                 {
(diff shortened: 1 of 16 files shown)


src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -40,6 +41,7 @@ public static IEnumerable<Example> Examples
             {
                 yield return new Example(Resources.Example_UpdateCommand_SearchAndUpdateVersionAndInstallerURL, new UpdateCommand { Id = "<PackageIdentifier>", InstallerUrls = new string[] { "<InstallerUrl1>", "<InstallerUrl2>" }, Version = "<Version>" });
                 yield return new Example(Resources.Example_UpdateCommand_SaveAndPublish, new UpdateCommand { Id = "<PackageIdentifier>", Version = "<Version>", OutputDir = "<OutputDirectory>", GitHubToken = "<GitHubPersonalAccessToken>" });
+                yield return new Example(Resources.Example_UpdateCommand_OverrideArchitecture, new UpdateCommand { Id = "<PackageIdentifier>", InstallerUrls = new string[] { "<InstallerUrl1>|<InstallerArchitecture>" }, Version = "<Version>" });
```
palenshus, line 44: Can you update the update readme.md too?

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -183,12 +183,14 @@ public static string GetFileHash(string path)
...
         /// <param name="newInstallers">List of newly parsed installer nodes.</param>
+        /// <param name="installerArchOverrideMap">Dictionary that maps the installer URL with the override architecture.</param>
         public static void UpdateInstallerNodesAsync(
             InstallerManifest installerManifest,
             IEnumerable<string> installerUrls,
             IEnumerable<string> paths,
             out List<DetectedArch> detectedArchOfInstallers,
-            out List<Installer> newInstallers)
+            out List<Installer> newInstallers,
+            Dictionary<string, InstallerArchitecture> installerArchOverrideMap)
```
palenshus, line 193: We now have three sort-of-arrays getting passed in that map to each other. How about we create a new input type with three properties, url, path, and overrideArch and pass in a list of those?

Optionally it could also have additional properties for Installer and DetectedArch and this method wouldn't need any out parameters, those things could just be set on the input objects themselves

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -334,6 +307,76 @@ public static Installer CloneInstaller(Installer installer)
...
+                    i => (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
+                    i.Architecture == detectedArch.OverrideArch);
+
+            if (overrideArchMatches.Count() == 1)
+            {
+                return overrideArchMatches.Single();
+            }
+
+            var urlArchMatches = existingInstallers.Where(
+                                        i => (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
```
palenshus, line 346: [nit]: the tabbing here is weird and different than line 337

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -334,6 +307,76 @@ public static Installer CloneInstaller(Installer installer)
...
+            ref List<Installer> multipleMatchedInstallers)
+        {
+            DetectedArch detectedArch = detectedArchOfInstallers.Find(
+                i => i.Url == newInstaller.InstallerUrl &&
+                (i.UrlArch == newInstaller.Architecture || i.BinaryArch == newInstaller.Architecture || i.OverrideArch == newInstaller.Architecture));
+
+            // If we can find an exact match by comparing the installerType and the override architecture, then return match.
+            // Otherwise, continue and try matching based on arch detected from url and binary detection, as the user could be trying overwrite with a new architecture.
+            var overrideArchMatches = existingInstallers.Where(
+                    i => (i.InstallerType ?? installerManifest.InstallerType) == newInstaller.InstallerType &&
```
palenshus, line 337: You can do this query once and save it to a temp variable, and then the three extra filter queries could all be based off of this, for example:

var overrideArchMatches archMatches.Where(i => i.Architecture == detectedArch.OverrideArch);

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -253,6 +253,119 @@ public async Task BlockUpdateSubmissionsWithNoUpdate()
             Assert.That(result, Does.Contain(Resources.NoChangeDetectedInUpdatedManifest_Message), "Failed to block manifests without updates from submitting.");
         }
 
+        /// <summary>
+        /// Verfies that an error message is shown if the overriding architecture is invalid.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
+        [Test]
```
palenshus, line 260: Nice tests!

palenshus (changes requested):
:clock1:

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: added example to readme

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
ryfu-msft: thanks :)

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: fixed weird tab spacing

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Changed to suggestion of only querying once and adding extra filters for the rest

src/WingetCreateCore/Models/InstallerUpdateHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.WingetCreateCore.Models
+{
+    using System.Collections.Generic;
+    using Microsoft.WingetCreateCore.Models.Installer;
+
+    /// <summary>
+    /// Helper class for storing information relating to updating an installer.
+    /// </summary>
+    public class InstallerUpdateHelper
```
palenshus, line 12: >InstallerUpdateHelper [](http://example.com/codeflow?start=17&length=21)

nit: I don't love this name. Helper to me implies something that exposes functionality to do things, this is really just a POCO class describing metadata about a url/installer. I think something like InstallerMetadata or InstallerUrlMetadata would be more apt?

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -220,32 +222,48 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
...
+            List<InstallerUpdateHelper> installerUpdateHelpers = this.ParseInstallerUrlsForArchOverride(this.InstallerUrls.ToList());
+
+            // If the installer update list is null there was an issue when parsing for architecture override.
+            if (installerUpdateHelpers == null)
+            {
+                return null;
+            }
+
+            // Reassign list with parsed installer urls.
+            this.InstallerUrls = installerUpdateHelpers.Select(x => x.InstallerUrl);
```
palenshus, line 235: How could this change from what's passed in?

palenshus, line 235: If you keep this, use .ToList() or .ToArray(). If it's just the IEnumerable, then it'll have to be re-evaluated on every access

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -419,6 +443,54 @@ private static void DisplayManifestsAsMenuSelection(Manifests manifests)
...
+                    }
+
+                    string installerUrl = installerUrlOverride[0];
+                    string overrideArchString = installerUrlOverride[1];
+                    InstallerArchitecture? overrideArch = overrideArchString.ToEnumOrDefault<InstallerArchitecture>();
+                    if (overrideArch.HasValue)
+                    {
+                        installerUpdate.InstallerUrl = installerUrl;
+                        installerUpdate.OverrideArchitecture = overrideArch.Value;
+                        installerUpdateList.Add(installerUpdate);
```
palenshus, line 476: Factor out of the if blocks

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -176,36 +160,28 @@ public static string GetFileHash(string path)
...
-                if (!ParsePackageAndGenerateInstallerNodes(path, url, newInstallers, null, ref detectedArchOfInstallers))
+                if (!ParsePackageAndGenerateInstallerNodes(installerUpdateHelper, null))
                 {
-                    parseFailedInstallerUrls.Add(url);
+                    parseFailedInstallerUrls.Add(installerUpdateHelper.InstallerUrl);
                 }
             }
 
+            List<Installer> newInstallers = new List<Installer>();
+            installerUpdateHelpers.ForEach(x => newInstallers.AddRange(x.NewInstallers));
```
palenshus, line 183: = installerUpdateHelpers.SelectMany(x => x.NewInstallers).ToList()

palenshus: Actually, if you're just wanting the count, you could do SelectMany(...).Count(), or even installerUpdateHelpers.Sum(x => x.NewInstallers.Count)

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -220,61 +196,36 @@ public static string GetFileHash(string path)
...
-                {
-                    installerMatchDict.Add(matchingExistingInstaller, newInstaller); // add the match to the map.
-                    existingInstallers.Remove(matchingExistingInstaller);
-                }
             }
 
             if (unmatchedInstallers.Any() || multipleMatchedInstallers.Any())
             {
-                throw new InstallerMatchException(multipleMatchedInstallers, unmatchedInstallers);
+                bool isArchitectureOverride = installerUpdateHelpers.Select(x => x.OverrideArchitecture).Any(i => i.HasValue);
```
palenshus, line 227: can just be installerUpdateHelpers.Any(x => x.OverrideArchitecture.HasValue)

palenshus (changes requested):
:clock1:

src/WingetCreateCore/Models/InstallerUpdateHelper.cs
ryfu-msft: Thanks for the suggestion, I was struggling to find a good name for this class. I changed the naming to "InstallerMetadata"

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: this needs to be reassigned because this.InstallerUrls will have the old installerUrl values with any architecture overrides appended by the user. ParseInstallerUrlsForArchOverride should return the actual installerUrls provided by the user, so this line is needed to make sure this.InstallerUrls is accurate.

ryfu-msft: Added .ToList() as per suggestion since I'm deciding to keep this line.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: factored out of if blocks

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: you are truly a linq magician. Thanks for the suggestion :) simplified to just determining the count of new installers.

src/WingetCreateCore/Common/PackageParser.cs
ryfu-msft: Thanks, changed to simplified suggestion

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: Thanks for adding the comment!

src/WingetCreateCLI/Commands/UpdateCommand.cs
palenshus: I meant the larger if blocks :)

ryfu-msft: ah i see, factored out of larger if block

palenshus (approved):
:shipit:

PR #215, opened 2022-01-19
Preserve existing ProductCode when updating

### LINKED ISSUE #212: ProductCode on installers is not preserved when updating existing manifest

I often post updates to the REAPER software to winget, which has `ProductCode: REAPER` set on both the x86 and x64 installers. However when running update with new urls and a new version, ProductCode is discarded in the new manifests.

### Steps to reproduce

wingetcreate update Cockos.REAPER -u https://www.reaper.fm/files/6.x/reaper645-install.exe https://www.reaper.fm/files/6.x/reaper645_x64-install.exe --version 6.45           

### Expected behavior

ProductCode is preserved twice in installer manifest

### Actual behavior

ProductCode is *bnot* preserved twice in installer manifest

### Environment

```shell
WingetCreateCLI 0.4.4.1
```

ryfu-msft:
Fixes #212 

Issue:
When updating an existingInstaller with a newInstaller, if the newInstaller does not have the productCode defined, it would overwrite the existing ProductCode. 

Fix:
In order to address this issue, I added a check to first see if the new ProductCode is non-null before assigning the value. That way the existing ProductCode can be preserved if no new value is present.

I have also applied the same logic to other relevant fields such as MinimumOSVersion, PackageFamilyName, and Platform.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/215)

### DIFF
--- a/src/WingetCreateCore/Common/PackageParser.cs
+++ b/src/WingetCreateCore/Common/PackageParser.cs
@@ -360,10 +360,12 @@ namespace Microsoft.WingetCreateCore
             existingInstaller.InstallerUrl = newInstaller.InstallerUrl;
             existingInstaller.InstallerSha256 = newInstaller.InstallerSha256;
             existingInstaller.SignatureSha256 = newInstaller.SignatureSha256;
-            existingInstaller.ProductCode = newInstaller.ProductCode;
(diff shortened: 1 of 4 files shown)


yao-msft (approved):
:shipit:

Issue #222, opened 2022-01-27
Wix not Differentiated from MSI


Trenly:
### Brief description of your issue

When a manifest with installer type 'wix' exists, attempting to update the package with new URLs fails as winget-create does not currently detect downloaded installers as wix

### Steps to reproduce

1) Find a package with installertype of wix
2) Attempt to update it with winget-create
3) get an error that url couldn't be matched to an installer

### Expected behavior

The URL will be matched to an existing WIX installer

### Actual behavior

```
wingetcreate update ChristianHohnstadt.xca -v 2.4.1 -u https://github.com/chris2511/xca/releases/download/RELEASE.2.4.0/xca-2.4.0.msi
Retrieving latest manifest for ChristianHohnstadt.xca
Downloading and parsing: https://github.com/chris2511/xca/releases/download/RELEASE.2.4.0/xca-2.4.0.msi...
Each new installer URL must have a single match to an existing installer based on installer type and architecture. The following installers failed to match an existing installer:
Multiple matches found for X64 Msi installer detected from the url: https://github.com/chris2511/xca/releases/download/RELEASE.2.4.0/xca-2.4.0.msi
```
\- provided by @OfficialEsco

### Environment

```shell
wingetcreate v0.4.4.1
```

PR #223, opened 2022-02-02
Remove notice file generation pipeline task

ryfu-msft:
Changes:
- Updates NOTICE.txt to latest
- Removes the NOTICE file generation pipeline task as there is a known issue with the Component Governance task when performing validation on forked PRs which will cause this step to fail. As seen in PR #222 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/223)

### DIFF
--- a/pipelines/azure-pipelines.yml
+++ b/pipelines/azure-pipelines.yml
@@ -113,28 +113,6 @@ jobs:
           verbosity: "Verbose"
           alertWarningLevel: "High"
 
-      - task: notice@0
-        displayName: "NOTICE File Generator"
-        inputs:
-          outputfile: "$(System.DefaultWorkingDirectory)/temp/NOTICE.txt"
-          outputformat: "text"
-
-      - bash: |
-          echo "Diffing existing NOTICE.txt with generated version"
-          diff -w NOTICE.txt temp/NOTICE.txt
-          if [[ $? -ne 0 ]];
-          then
-            echo "Notice file modified"
-            echo "*******************************************************************************************************"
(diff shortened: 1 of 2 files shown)


pipelines/azure-pipelines.yml
```diff
@@ -113,28 +113,6 @@ jobs:
           verbosity: "Verbose"
           alertWarningLevel: "High"
 
-      - task: notice@0
```
yao-msft, line 116: Normally we would temporarily disable this task for pr validation pipeline by adding a condition on the task. Then when Component Governance team fixed the issue after the follow up, reenable the task.
https://github.com/microsoft/winget-cli/pull/775/files

ryfu-msft: I decided to make the task continue on error. The reason is that this only fails on forked PRs, and would still work ona branch from the source repo. That way we can still update the NOTICE.txt file every so often, and if the CG team is able to fix the issue, it would also work for forked PRs.

pipelines/azure-pipelines.yml
```diff
@@ -115,6 +115,7 @@ jobs:
 
       - task: notice@0
         displayName: "NOTICE File Generator"
+        continueOnError: "true"
```
yao-msft, line 118: >continueOnError: "true" [](http://example.com/codeflow?start=8&length=23)

condition: not(eq(variables['Build.Reason'], 'PullRequest'))

ryfu-msft: changed to suggested

yao-msft (approved):

PR #224, opened 2022-02-09
Add UpdateCommand support for 1.1 Schema

### LINKED ISSUE #139: Support Windows Package Manager v1.1 schema

The tool should support the new v1.1 schema. It should also be able to update v1.0 manifests to v1.1.

ryfu-msft:
Resolves #139 

Changes:
- Increment version from 0.4.4 to 0.5 in preparation for new release
- Adds a custom AliasTypeInspector that prioritizes properties with the YamlMemberAttribute to handle custom properties overrides such ReleaseDate and Agreement.
- Updates the Automapper to include mapping for for newly added object models (i.e. AppsAndFeaturesEntry, ExpectedReturnCode, PackageDependencies, Markets, etc.)
- Ensures UpdateCommand updates ManifestVersion field to the latest schema version.

Verification:
Added 3 unit tests to cover the above changes
- UpdatesToLatestManifestVersion()  
- UpdateDeserializesAliasDefinedFields()
- UpdateFullSingletonVersion1_1()

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/224)

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -360,6 +360,12 @@ namespace Microsoft.WingetCreateCLI.Commands
                 cfg.CreateMap<WingetCreateCore.Models.Singleton.Dependencies, WingetCreateCore.Models.Installer.Dependencies>();
                 cfg.CreateMap<WingetCreateCore.Models.Singleton.Installer, WingetCreateCore.Models.Installer.Installer>();
                 cfg.CreateMap<WingetCreateCore.Models.Singleton.InstallerSwitches, WingetCreateCore.Models.Installer.InstallerSwitches>();
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.AppsAndFeaturesEntry, WingetCreateCore.Models.Installer.AppsAndFeaturesEntry>();
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.ExpectedReturnCode, WingetCreateCore.Models.Installer.ExpectedReturnCode>();
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.PackageDependencies, WingetCreateCore.Models.Installer.PackageDependencies>();
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.Markets, WingetCreateCore.Models.Installer.Markets>();
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.Markets2, WingetCreateCore.Models.Installer.Markets2>(); // Markets2 is not used, but is required to satisfy mapping configuration.
+                cfg.CreateMap<WingetCreateCore.Models.Singleton.Agreement, WingetCreateCore.Models.DefaultLocale.Agreement>();
             });
             var mapper = config.CreateMapper();
 
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -24,7 +24,6 @@ namespace Microsoft.WingetCreateCLI.Commands
     using Microsoft.WingetCreateCore.Models.Locale;
     using Microsoft.WingetCreateCore.Models.Version;
     using Sharprompt;
-    using Windows.Media.SpeechRecognition;
 
     /// <summary>
     /// Command for updating the elements of an existing or local manifest.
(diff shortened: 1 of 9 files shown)


src/WingetCreateCore/Common/Serialization.cs
```diff
@@ -197,18 +199,61 @@ private static string RemoveBom(string value)
...
+
+            /// <summary>
+            /// Because certain properties were generated incorrectly, we needed to create custom fields for those properties.
+            /// Therefore to resolve naming conflicts during deserialization, we prioritize fields that have the YamlMemberAttribute defined
+            /// as that attribute indicates an override.
+            /// </summary>
+            public override IEnumerable<IPropertyDescriptor> GetProperties(Type type, object container)
+            {
+                var propertyDescriptors = this.innerTypeDescriptor.GetProperties(type, container);
+                var aliasDefinedProps = type.GetProperties().ToList().Where(p => p.GetCustomAttribute<YamlMemberAttribute>() != null).ToList();
```
yao-msft, line 224: >aliasDefinedProps [](http://example.com/codeflow?start=20&length=17)

nit: yamlMemberAttributeDefinedProps

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -313,6 +313,8 @@ public Manifests DeserializeManifestContentAndApplyInitialUpdate(List<string> la
                 manifests = ConvertSingletonToMultifileManifest(manifests.SingletonManifest);
             }
 
+            UpdateManifestVersionToLatest(manifests);
```
yao-msft, line 316: >UpdateManifestVersionToLatest [](http://example.com/codeflow?start=12&length=29)

Curious why is update always updating to latest schema version? Is there an option to a specific schema version?

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -379,6 +390,18 @@ private static void UpdatePropertyForLocaleManifests(string propertyName, string
             }
         }
 
+        /// <summary>
+        /// Ensures that the manifestVersion is updated to the latest schema version.
+        /// </summary>
+        /// <param name="manifests">Manifests object model.</param>
+        private static void UpdateManifestVersionToLatest(Manifests manifests)
```
yao-msft, line 397: >UpdateManifestVersionToLatest [](http://example.com/codeflow?start=28&length=29)

nit: EnsureManifestVersionConsistency

src/WingetCreateCLI/Commands/UpdateCommand.cs
denelon: We want to keep manifests as current as reasonable. We've been discussing enabling an experimental feature to try newer versions of the schema early. I'm not sure why we would want to support an older schema other than the scenario where a user is running an older version of a REST source.

yao-msft: Keeping schema version as latest as possible is probably ok for v1.0 and v1.1 schemas because they are compatible. But if we have a breaking schema version like v2.0(given we still do not drop v1.0 schemas), this will be an issue. For example if a user calls "winget-create update <SomeV1.0Manifest>" to maybe just update some description fields or installer url of a manifest, suddenly the manifest becomes a v2.0 manifest and the package lost support from old clients where only v1.0 is recognized. We may certainly encourage latest schema version, but I think we should still support schema versions  that are not deprecated yet.

src/WingetCreateCore/Common/Serialization.cs
ryfu-msft: Added additional check to verify that alias is defined. Therefore I'm keeping the name because it still applies.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Changed to suggested.

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: I have created a new feature issue to track this here: #225

src/WingetCreateCLI/Commands/UpdateCommand.cs
yao-msft: Forgot to update the summary..

yao-msft (approved):
:shipit:

PR #230, opened 2022-03-01
Allow installerType to be updated with more precise, compatible installerType

### LINKED ISSUE #227: Allow change of installer type MSI => WIX

There are several packages in the community repository with the installer type currently set to MSI incorrectly. Since the addition of #217 and #220, wingetcreate can now detect when an MSI file is actually built with WIX (thanks @jedieaston!) and will match the WIX installer to the MSI entry. However, this relies on the user to update the installer type themselves, even if wingetcreate can detect it reliably. 

As a user, it would be nice if installer types would automatically be updated if the detected installer type is compatible with and more specific than the installer type already defined in the manifest. As an example, InstallerType of MSI could be changed to WIX, but WIX should not be changed to MSI. Similarly, EXE could be changed to INNO, NULLSOFT, or BURN, but not the other way around;

_No response_

ryfu-msft:
Resolves #227 

Currently, WingetCreate's Update command only checks if two installerTypes are compatible when matching installers. However, the manifest does not get updated with the more accurate/precise installerType. i.e. Msi -> Wix.

Changes:
- Added logic that checks if an installerType is more precise. If so, the update will overwrite the existing InstallerType value with the more precise, compatible installerType value. 

Tests:
- Added unit test UpdateReplacesWithMorePreciseInstallerType() that verifies with a wix installer that the manifest gets updated from MSI -> WIX.


src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -370,7 +370,12 @@ public static Installer CloneInstaller(Installer installer)
         /// <param name="existingInstaller">Existing installer node.</param>
         /// <param name="newInstaller">New installer node.</param>
         private static void UpdateInstallerMetadata(Installer existingInstaller, Installer newInstaller)
-        {
+        {
+            if (IsPreciseInstallerType(newInstaller.InstallerType.Value))
+            {
+                existingInstaller.InstallerType = newInstaller.InstallerType;
```
yao-msft, line 376: >existingInstaller.InstallerType = newInstaller.InstallerType; [](http://example.com/codeflow?start=16&length=61)

This has more implication than simply overriding the installer type. It might break installer if it's not done correctly. For example, Installer Switches might not be same for different installer types. I suggest we revisit Installer fields to see if some fields need to be reset, we need to also consider what the impact will be for root level fields.

ryfu-msft:
I am closing this PR for now as this will require further investigation regarding how installer switches are transferred between compatible installer types and whether some installer fields need to be reset in order to safely update the installer type.

PR #257, opened 2022-04-09
Same filename

### LINKED ISSUE #249: Incorrect hashes computed when installer file names are not unique

When multiple installer URLs are passed that use a common name, winget-create overwrites the downloaded file in the temp directory and subsequently computes the same hash for the different variants.

### Steps to reproduce

Run
```
wingetcreate update WillowSoftware.AnvilStudio -v 22.03.02 -u https://anvilstudio.com/x/asinstall.exe https://anvilstudio.com/x64/asinstall.exe
```

### Expected behavior

The generated manifest should contain the correct file hashes for the two installer variants.

### Actual behavior

The generated manifest contains the wrong file hash for the x86 variant:
```yaml
Installers:
- Architecture: x64
  InstallerType: exe
  InstallerUrl: https://anvilstudio.com/x64/asinstall.exe
  InstallerSha256: 02FC70CD7CCC165A1F9089F0EF29BD3BA2E1A0374D502A7E7784E7BA5C21AE46
- Architecture: x86
  InstallerType: exe
  InstallerUrl: https://anvilstudio.com/x/asinstall.exe
  InstallerSha256: 02FC70CD7CCC165A1F9089F0EF29BD3BA2E1A0374D502A7E7784E7BA5C21AE46
```

### Environment

```shell
Windows Manifest Creator v1.0.3.0
```

Trenly:
  - Resolves #249

This causes files with the same name to be downloaded with the numeric filenames that Windows uses elsewhere

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/257)

### DIFF
--- a/src/WingetCreateCore/Common/PackageParser.cs
+++ b/src/WingetCreateCore/Common/PackageParser.cs
@@ -147,6 +147,19 @@ namespace Microsoft.WingetCreateCore
 
             using var targetFileStream = File.OpenWrite(targetFile);
             var contentStream = await response.Content.ReadAsStreamAsync();
+
+            /*
+             * There seems to be a difference between the test environments and a users environment
+             * On the users environment, the stream cannot seek and will always have position 0 when
+             * the response content is read into it. In the unit test environment, the stream can
+             * seek and will not reset the position when the response content is read in. This logic
+             * should always ensure that the position is reset to 0 regardless of the environment.
+            */
+            if (contentStream.CanSeek)
+            {
+                contentStream.Position = 0;
+            }
+
             await contentStream.CopyToAsync(targetFileStream);
 
             return targetFile;
--- a/src/WingetCreateCore/Common/PackageParser.cs
+++ b/src/WingetCreateCore/Common/PackageParser.cs
@@ -128,6 +128,13 @@ namespace Microsoft.WingetCreateCore
                 throw new HttpRequestException(message, null, response.StatusCode);
             }
 
+            long? downloadSize = response.Content.Headers.ContentLength;
+
+            if (downloadSize > maxDownloadSize)
+            {
+                throw new DownloadSizeExceededException(maxDownloadSize.Value);
+            }
+
             string urlFile = Path.GetFileName(url.Split('?').Last());
             string contentDispositionFile = response.Content.Headers.ContentDisposition?.FileName?.Trim('"');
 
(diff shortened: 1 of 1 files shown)


Trenly:
@ryfu-msft I can't seem to figure out why it is failing the `UpdateWithArchitectureOverrides` test. I'm not seeing anywhere in the code that it would depend on the path to the downloaded file in the tests, unless I'm overlooking it?

ryfu-msft:
Okay, after doing some digging I think I found the issue. When I run the test for the first time (clearing out any previously downloaded test installers), I see this:

image

For the `UpdateWithArchitectureOverride`, we actually run the update command twice, once without the architecture override, and once with the architecture override. Because of this, we should expect the `DownloadFileAsync()` function to be called twice. I found it odd that the 2nd instance always produced a file with 0KB, which is why the error message shows something weird like `Unable to read beyond the end of the stream`.

Based on this [post](https://stackoverflow.com/questions/60775457/stream-copytoasync-is-empty-after-first-iteration), it appears that after copying from a stream, the pointer is left at the last position and needs to be rewind back to 0. I think that should help resolve the issue. 

``` C#
            using var targetFileStream = File.OpenWrite(targetFile);
            var contentStream = await response.Content.ReadAsStreamAsync();
            contentStream.Position = 0;  
            await contentStream.CopyToAsync(targetFileStream);
````

File: src/WingetCreateCore/Common/PackageParser.cs [hunk 1]
```diff
@@ -144,14 +144,19 @@ public static async Task<string> DownloadFileAsync(string url, long? maxDownload
...
             {
-                File.Delete(targetFile);
-                using var targetFileStream = File.OpenWrite(targetFile);
-                var contentStream = await response.Content.ReadAsStreamAsync();
-                await contentStream.CopyToAsync(targetFileStream);
+                string fileName = Path.GetFileNameWithoutExtension(targetFile);
+                string fileExt = Path.GetExtension(targetFile);
+                DirectoryInfo dir = new DirectoryInfo(InstallerDownloadPath);
+                FileInfo[] existingFiles = dir.GetFiles(Path.ChangeExtension(fileName + " *", fileExt));
+                targetFile = Path.Combine(InstallerDownloadPath, fileName + " (" + (existingFiles.Length + 1).ToString() + ")" + fileExt);
```
ryfu-msft, line 153: Can you separate this out into it's own function so it's easier to read and can be used elsewhere if needed? Maybe like `GetNumericFilename()`

ryfu-msft (changes requested):

Trenly:
> Okay, after doing some digging I think I found the issue. When I ran the test for the first time (clearing out any previously downloaded test installers), I see this:
> 
> image
> 
> For the `UpdateWithArchitectureOverride`, we actually run the update command twice, once without the architecture override, and once with the architecture override. Because of this, we should expect the `DownloadFileAsync()` function to be called twice. I found it odd that the 2nd instance always produced a file with 0KB, which is why the error message shows something weird like `Unable to read beyond the end of the stream`.
> 
> Based on this [post](https://stackoverflow.com/questions/60775457/stream-copytoasync-is-empty-after-first-iteration), it appears that after copying from a stream, the pointer is left at the last position and needs to be rewind back to 0. I think that should help resolve the issue.
> 
> ```cs
>             using var targetFileStream = File.OpenWrite(targetFile);
>             var contentStream = await response.Content.ReadAsStreamAsync();
>             contentStream.Position = 0;  
>             await contentStream.CopyToAsync(targetFileStream);
> ```

So, that makes the tests pass, but it also causes this error when actually running the client - 
```
Failed to download installer.
System.NotSupportedException: Specified method is not supported.
   at System.Net.Http.HttpBaseStream.set_Position(Int64 value)
   at Microsoft.WingetCreateCore.PackageParser.DownloadFileAsync(String url, Nullable`1 maxDownloadSize)
   at Microsoft.WingetCreateCLI.Commands.BaseCommand.DownloadPackageFile(String installerUrl)
   at Microsoft.WingetCreateCLI.Commands.UpdateCommand.UpdateManifestsAutonomously(Manifests manifests)
```

Trenly:
> Azure Pipelines successfully started running 1 pipeline(s).

```
##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider InternalServerError
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider InternalServerError
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider InternalServerError
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider InternalServerError
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider InternalServerError
,##[error]Provisioning request delayed or failed to send 5 time(s). This is over the limit of 3 time(s).
```

ryfu-msft:
Hmm, weird... looking into this.

Trenly:
I wonder if it has something to do with the commit history? I can try rebasing and force pushing

Trenly:
Hmm. Thats weird

Trenly:
Now it's trying to run it on https://github.com/microsoft/winget-create/commit/6ebb719c43926a72712bce48d49c40b2f49a94a6 which isn't the latest commit

Trenly:
Trying in #262

PR #235, opened 2022-03-08
Apply fast-forward update to default branch of forked repo prior to submission

ryfu-msft:
Resolves #233 
Resolves #234 

GitHub API will throw a NotFoundException when trying to create a reference if the default branch of the forked 'winget-pkgs' repository is not up to date with the latest upstream commits. 

The changes in this PR resolve these issues by checking for any upstream commits and updating the fork's default branch prior to submission. The update will only work successfully if the update is only a fast-forward update.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/235)

PR #289, opened 2022-07-22
Sync fork prior to creating reference and handle NotFoundException

### LINKED ISSUE #287: NotFoundException error during pull request creation

There is a known issue where winget-create is unable to create a reference to the repository if the main/master branch is behind by a certain number of commits. To resolve this issue, the user needs to perform the following steps to update the master branch:

```
git remote add upstream git@github.com:microsoft/winget-pkgs.git
git switch master
git reset upstream/master --hard 
git push
```

### Steps to reproduce

Leave the master branch of a forked winget-pkgs repository unchanged until it is behind the upstream/master branch by a large number of commits.

### Expected behavior

Pull request should be created

### Actual behavior

NotFoundException is thrown and pull request does not get created

### Environment

```shell
Winget-Create 1.0.4.0
```

ryfu-msft:
Fixes #287 

A NotFoundException occurs when the forked repo is behind by too many commits. This fails even when attempting to sync the forked repo with the upstream which nullifies my previous fix #235 

To address this issue, I have added the following changes:
1. Always attempt to sync fork but make it nonblocking if it fails as creating a reference could still possibly succeed. (Example: unable to sync fork because it is a fast-forward update) 
2. Add a friendly error message if a NotFoundException is encountered instructing the user to manually sync their fork. 

Current tests should verify that submission succeeds as this is a minor edge case that is difficult to replicate due to the requirement of the fork needing to be behind by X amount of commits.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/289)

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -503,9 +503,11 @@ namespace Microsoft.WingetCreateCLI.Commands
                     Logger.ErrorLocalized(nameof(Resources.Error_Prefix), e.Message);
                     return false;
                 }
-                else if (e is NonFastForwardException nonFastForwardException)
+                else if (e is Octokit.NotFoundException)
                 {
-                    Logger.ErrorLocalized(nameof(Resources.FastForwardUpdateFailed_Message), nonFastForwardException.CommitsAheadBy);
+                    // This exception can occur if the client is unable to create a reference due to being behind by too many commits.
+                    // The user will need to manually update their master branch of their winget-pkgs fork.
+                    Logger.ErrorLocalized(nameof(Resources.SyncForkWithUpstream_Message));
                     return false;
                 }
                 else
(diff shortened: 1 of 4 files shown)


Trenly:
I'm not sure how winget-create handles the forks and branches, but given that we have the upstream information in order to update the fork, couldn't the commit be applied directly to a detached head which is the latest commit to the default branch of upstream, removing the need to check if the fork is behind at all?

ryfu-msft:
> I'm not sure how winget-create handles the forks and branches, but given that we have the upstream information in order to update the fork, couldn't the commit be applied directly to a detached head which is the latest commit to the default branch of upstream, removing the need to check if the fork is behind at all?

Linking to the source code in case you were curious how it was implemented: https://github.com/microsoft/winget-create/blob/49a1fb42f5b7f802d920207b6b48176670b7f6d4/src/WingetCreateCore/Common/GitHub.cs#L278-L378

We already assign the new reference with the latest SHA1 value directly from the upstream default branch before attempting to create it. Even though we don't interact with the forked repo except when creating the new branch/reference, it seems like an outdated fork unintentionally impacts the `Reference.Create()` function and causes it to fail.

yao-msft (approved):

PR #293, opened 2022-08-03
Fix null/empty filename string when downloading package file

### LINKED ISSUE #292: Installer URIs being a redirect cause a `System.UnauthorizedAccessException`

When I supply an Installer URI that is a redirect to the actual download link, I get the `System.UnauthorizedAccessException` as shown below.

### Steps to reproduce

1. run `wingetcreate new`
1. Supply `https://www.two-notes.com/download/76391/` as installer url (it is the Windows / x64 installer from https://www.two-notes.com/en/torpedo-series/torpedo-remote/)

### Expected behavior

winget-create follows the redirect and downloads the installer from the correct path.

### Actual behavior

```
Downloading and parsing: https://www.two-notes.com/download/76391/...
Failed to download installer.
System.UnauthorizedAccessException: Access to the path 'C:\Users\HansBakker\AppData\Local\Temp\wingetcreate' is denied.
   at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at Microsoft.WingetCreateCore.PackageParser.DownloadFileAsync(String url, Nullable`1 maxDownloadSize)
   at Microsoft.WingetCreateCLI.Commands.BaseCommand.DownloadPackageFile(String installerUrl)
   at Microsoft.WingetCreateCLI.Commands.NewCommand.Execute()
```
image

### Environment

```shell
Windows Package Manager v1.2.11601
Windows: Windows.Desktop v10.0.22000.795
Package: Microsoft.DesktopAppInstaller v1.17.11601.0
```

ryfu-msft:
Fixes #292 

The changes in this PR fix an issue where the target filename used to create the package file is null. This can cause an UnauthorizedAccessException since an empty filename means that the File.OpenWrite is trying to write to a directory instead of a file. To resolve this, I use the filename from the requestUrl as the last option for the filename in case both the contentDispositionFile and urlFile names are both empty. 

Verified manually with the related issue.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/293)

### DIFF
--- a/src/WingetCreateCore/Common/Extensions.cs
+++ b/src/WingetCreateCore/Common/Extensions.cs
@@ -96,6 +96,16 @@ namespace Microsoft.WingetCreateCore.Common
         }
 
         /// <summary>
+        /// Returns null if the target string is empty.
+        /// </summary>
+        /// <param name="s">Target string value.</param>
+        /// <returns>Null if the string is empty; otherwise the original string value.</returns>
+        public static string NullIfEmpty(this string s)
+        {
+            return string.IsNullOrEmpty(s) ? null : s;
+        }
+
+        /// <summary>
         /// Determines if the properties of an object are all equal to null excluding properties with dictionary type if needed.
         /// </summary>
         /// <param name="o">Object to be evaluated.</param>
--- a/src/WingetCreateCore/Common/PackageParser.cs
+++ b/src/WingetCreateCore/Common/PackageParser.cs
@@ -136,14 +136,15 @@ namespace Microsoft.WingetCreateCore
 
             string urlFile = Path.GetFileName(url.Split('?').Last());
             string contentDispositionFile = response.Content.Headers.ContentDisposition?.FileName?.Trim('"');
+            string requestUrl = response.RequestMessage.RequestUri.ToString();
 
             if (!Directory.Exists(InstallerDownloadPath))
             {
                 Directory.CreateDirectory(InstallerDownloadPath);
             }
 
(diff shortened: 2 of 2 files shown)


src/WingetCreateCore/Common/Extensions.cs
```diff
@@ -93,6 +93,16 @@ public static string ToEnumAttributeValue(this Enum enumVal)
...
             EnumMemberAttribute attributeValue = (attributes.Length > 0) ? (EnumMemberAttribute)attributes[0] : null;
             return attributeValue?.Value ?? enumVal.ToString();
+        }
+
+        /// <summary>
+        /// Returns null if the target string is empty.
+        /// </summary>
+        /// <param name="s">Target string value.</param>
+        /// <returns>Null if the string is empty; otherwise the original string value.</returns>
+        public static string NullIfEmpty(this string s)
```
yao-msft, line 103: Is this common practice?

yao-msft (approved):

src/WingetCreateCore/Common/Extensions.cs
ryfu-msft: I believe it is, since it is commonly used with the null coalescing operator when handling empty strings

PR #330, opened 2022-12-07
Display error message during network connection failure.

### LINKED ISSUE #328: wingetcreate does not provide error when it cannot connect to server

When the network is unavailable, wingetcreate will silently error and the user does not know why.
This is related to issue #327

### Steps to reproduce

turn off network capability
type wingetcreate new
notice you  are returned to prompt and nothing happens.

C:\Users\kevinla>wingetcreate new
C:\Users\kevinla>

### Expected behavior

If there is an error due to lack of network, the user should be informed.
"Cannot connect to foo.  check your network connections and try again."

### Actual behavior

C:\Users\kevinla>wingetcreate new

C:\Users\kevinla>

### Environment

```shell
C:\Users\kevinla>wingetcreate
Windows Package Manager Manifest Creator v1.1.2.0
```

ryfu-msft:
Fixes #328 

Shows an explicit error message to the user if winget-create is unable to connect to GitHub when there is no network connection. The error is non-blocking in case connection to GitHub fails, but winget-create is still able to download from a local network.

Also displays the proper error message if GitHub oauth device login flow fails also fails to connect.

Verified behavior manually.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/330)

### DIFF
--- a/src/WingetCreateCLI/Program.cs
+++ b/src/WingetCreateCLI/Program.cs
@@ -5,7 +5,6 @@ namespace Microsoft.WingetCreateCLI
 {
     using System;
     using System.Linq;
-    using System.Net.Http;
     using System.Threading.Tasks;
     using CommandLine;
     using CommandLine.Text;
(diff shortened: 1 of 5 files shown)


sophacola (approved):

sophacola (approved):

src/WingetCreateCLI/Program.cs
```diff
@@ -71,7 +71,11 @@ private static async Task<int> Main(string[] args)
                 }
                 else
                 {
-                    return 1;
+                    // Do not block creating a new manifest if loading the GitHub client fails. InstallerURL could point to a local network.
```
yao-msft, line 74: Is local network case common? Will this expose more issues than it solves for some uncommon cases?

yao-msft (approved):

Issue #359, opened 2023-04-13
Add ability to pass in a PR title with submit and update


edwardskrod:
## What 
Add ability to change the automatically-generated PR title when submitting PR.

## Why
When submitting a PR with WinGetCreate update or WinGetCreate submit, the PR titles are automatically generated from the PackageIdentifier <id> and PackageVersion -v <version>
There are times when it would be nice to update the PR title with a custom PR title.

For example, Visual Studio has Preview versions that get released on a regular cadence.   Visual Studio 17.6 has 8 Preview Versions.  

Preview versions are based on the first release of the minor, i.e. 17.6.0.  

Right now, each preview version's PR in GitHub is identical:  

Microsoft.VisualStudio.2022.Community.Preview version 17.6.0

Instead, Visual Studio would like to be able to add the Preview version to the title.

- Microsoft.VisualStudio.2022.Community.Preview version 17.6.0 Preview 1
- Microsoft.VisualStudio.2022.Community.Preview version 17.6.0 Preview 2
- Microsoft.VisualStudio.2022.Community.Preview version 17.6.0 Preview 3

To do this today, we'd need to update the version from "17.6.0" to "17.6.0 Preview 1".  But this would also update the package version in the manifest, which is something we don't want to do.

Instead, we propose to allow WinGetCreate to accept a parameter to override the PR title.

### Proposed technical implementation details

wingetcreate.exe submit [\<options>] <PathToManifest> [-p \<pr title>]
wingetcreate.exe update \<id> [-u \<urls>] [-v \<version>] [-s] [-t \<token>] [-o \<output directory>] [-p \<pr title>]

Accept the PR title as a command line parameter for the submit and update commands.  Use this title rather than automagically generating the title in code from the Id and Version..

ryfu-msft:
Fixed in PR #362

PR #362, opened 2023-04-17
Add arg for changing pull request title

mdanish-kh:
    - #359

### Description of the changes
- Add an optional parameter `prTitle` to BaseCommand.GitHubSubmitManifests and avoid breaking existing calls of this function.
- For PR Title validation, let GitHub Api handle the validation and show exception message to user. The downside is (I think) the messages won't be localized, but the upside being anytime the validation (let's say PR title max length) changes at the api side, we wouldn't have to update it manually in the code 👀
 - In case title exceeds max length, it will be caught in `ApiValidationException`. The current limit seems to be 1024 characters (albeit the error response message says 256 chars)
   - [image.png (error response)](https://user-images.githubusercontent.com/88161975/232553008-e1294be6-14f5-4b5e-ae34-754675776046.png) 
   - [image2.png (exception handled in winget-create)](https://user-images.githubusercontent.com/88161975/232553949-dfa4b6b3-a4b9-4dcd-b84c-4d163c2fd274.png)
- If no title is provided, the default message of `<PackageId> version <PackageVersion>` will be used. 
- For empty strings with only whitespaces, the exception is caught in `ArgumentException`
  - [image.png (error response)](https://user-images.githubusercontent.com/88161975/232555161-dd5ecb5d-45c6-4d9b-99f4-c0334b8899e2.png)
  - [image2.png (exception handled in winget-create)](https://user-images.githubusercontent.com/88161975/232555635-fd3d696b-f1b7-4a7c-aba1-74d6291a115c.png)
- Updated E2E tests and docs for the new parameter

### Need feedback
Since `--prtitle` for `update` command is only applicable when the `--submit` flag is used, then what if a user was to use --prtitle in the update command without setting the submit flag?

I looked at getting a native solution from CommandLineParser to handle this option dependency on another option, but couldn't find the capability in the module to do so. Related issue: https://github.com/commandlineparser/commandline/issues/3

The alternative would be to handle it manually in the code and print a localized warning message something like "Submit flag is not set. The value of title is ignored.", but since this seems like a harmless scenario I decided not to handle it. Please let me know if I am thinking wrong about any of this.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/362)


ryfu-msft (approved):
I think what you have is perfectly fine. Ideally, we would want to show a warning message to the user that --submit is needed in combination with --prtitle, but I think the user will ultimately figure that out when they realize that no PR was submitted. I agree with you that this is a pretty harmless/trivial scenario. Everything else looks good :)

mdanish-kh:
Oopsies, found an unwanted change I mistakenly made in `Resources.Designer.cs`. Fixed in the recent commit.

Issue #371, opened 2023-05-10
Multiple nested installers are handled improperly


mdanish-kh:
### Brief description of your issue

The current master build has a very odd behavior in case of multiple nested installers where it prompts for PortableCommandAlias `nestedInstallers.Count` number of times and creates an incorrect manifest.
![image](https://github.com/microsoft/winget-create/assets/88161975/66fc440f-322a-47ff-bc8d-9eca945be39c)

Even if you select separate architecture for each of the nested installer, the final manifest will contain only a single installer because of how winget-create handles nested installers in the code (if you change one property of a nested installer object, it'll get modified in all of them so the final architecture selected by the user will be propagated in all installers)

### Steps to reproduce

1. Select same architecture for all of the nested installers
2. Select different architectures for all of the nested installers

#### Examples 

From [Microsoft.Sysinternals.ZoomIt](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/Sysinternals/ZoomIt/6.12/Microsoft.Sysinternals.ZoomIt.installer.yaml) (3 nested portables all with **different** architecture)
```
wingetcreate new https://download.sysinternals.com/files/ZoomIt.zip
```

From [PyPy.PyPy.3.9](https://github.com/microsoft/winget-pkgs/blob/master/manifests/p/PyPy/PyPy/3/9/7.3.11/PyPy.PyPy.3.9.installer.yaml) (4 nested portables all with **same** architecture)
```
wingetcreate new https://downloads.python.org/pypy/pypy3.9-v7.3.11-win64.zip
```

From [yelosan.PH-UKL.TglgQRTY](https://github.com/microsoft/winget-pkgs/blob/master/manifests/y/yelosan/PH-UKL/TglgQRTY/2.0.1/yelosan.PH-UKL.TglgQRTY.installer.yaml) (17 nested installers of different InstallerTypes i.e., portable, MSI)
```
wingetcreate new https://codeberg.org/attachments/7adf84ee-84a4-41d7-adb0-9f2d4aa979d5
```

From [Git.MinGit](https://github.com/microsoft/winget-pkgs/blob/master/manifests/g/Git/MinGit/2.40.1/Git.MinGit.installer.yaml) (two URLs, same arch for each URL)
```
[...]

PR #372, opened 2023-05-11
Improve handling of multiple nested installers

mdanish-kh:
   - #371

### Brief Description

1. Allocate a new installer object for each nested installer so that changes in one nested installer object do not propagate into another
2. As a consequence of the above change, logic needs to be added for merging nested installer files into a single installer in case of `NestedInstallerType: portable`. This is done as:
   * If two (or more) nested portable installers have the same architecture and hash, the only difference would be of RelativeFilePath in both of them; so we merge the two installers
3. Display name of NestedInstallerFile for better clarity before a prompt asking for additional metadata or asking whether the package is a portable.
4. Refactor PortableCommandAliasIfApplicable() as we expect a single NestedInstallerFiles node for each installer inside the method body now.

I've listed some test URLs/manifests in the linked issue

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/372)

### DIFF
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -172,15 +172,18 @@ namespace Microsoft.WingetCreateCLI.Commands
                             selectedInstallers = Prompt.MultiSelect(Resources.SelectInstallersFromZip_Message, extractedFiles, minimum: 1).ToList();
                         }
 
-                        installerUpdateList.Add(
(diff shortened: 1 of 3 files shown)


ryfu-msft (approved):
Tested manually with the examples in the bug issue thread. Implementation looks good. Thanks a ton! 
:shipit:

PR #388, opened 2023-06-17
Use common pull request titles

### LINKED ISSUE #387: Standard Pull Request Titles

Other [manifest authoring tools](https://github.com/microsoft/winget-pkgs#authoring-a-manifest) set a default PR title depending on the nature of the PR.

These titles help moderators reviewing packages know before-hand what to expect from the PR (i.e., If it's a new package we might want to check that the package doesn't already exist under some other PackageIdentifier, if it's an update we may skip this check) 

New package added in repo: `New package: <PackageId> version <Version>` [Examples](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+in%3Atitle+New+package%3A+is%3Aclosed+)
Update to a newer version: `New version: <PackageId> version <Version>` [Examples](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+is%3Aclosed+in%3Atitle+New+version)
Update to the existing version: `Update version: <PackageId> version <Version>` [Examples](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+is%3Aclosed+in%3Atitle+Update%3A)
Adding an older version to repo: `Add version: <PackageId> version <Version>` [Examples](https://github.com/microsoft/winget-pkgs/pulls?q=is%3Apr+in%3Atitle+Add+version%3A+is%3Aclosed+)

mdanish-kh:
   - Resolves #387

I also modified the default branch format to help find the branch more easily through `git checkout <>` (Tab completion) / `git branch -a`. This helps when a mod requests changes on a PR and the author wants to find the local branch and add changes

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/winget-create/pull/388)


src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -250,7 +250,9 @@ public override async Task<bool> Execute()
                 {
                     if (await this.LoadGitHubClient(true))
                     {
-                        return commandEvent.IsSuccessful = await this.GitHubSubmitManifests(manifests);
+                        return commandEvent.IsSuccessful = await this.GitHubSubmitManifests(
+                            manifests,
+                            $"New package: {manifests.InstallerManifest.PackageIdentifier} version {manifests.InstallerManifest.PackageVersion}");
```
Trenly, line 255: I don't think this is always accurate, since the `wingetcreate new` command can also be used to add new versions of an existing package

mdanish-kh: Perhaps you're confusing YamlCreate's behavior with wingetcreate? `wingetcreate new` will fail if it detects a PackageId that already exists in the repo
image

Trenly: > Perhaps you're confusing YamlCreate's behavior with wingetcreate? `wingetcreate new` will fail if it detects a PackageId that already exists in the repo image

Really? That's interesting, I could have sworn I saw a few PR's before where `wingetcreate new` was used to create an update

mdanish-kh: Am I missing something here? 😅 How did you know if the package was made with `wingetcreate new` or `wingetcreate update` from the PR title/contents/schema header? Cause wingetcreate doesn't really give any hints about that by default.

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -735,5 +737,24 @@ private async Task UpdateSingleInstallerInteractively(Installer installer)
...
+            // Handle case where manifest is still using singleton format
+            string repositoryVersion = repositoryManifest.SingletonManifest != null ? repositoryManifest.SingletonManifest.PackageVersion : repositoryManifest.InstallerManifest.PackageVersion;
+
+            return WinGetUtil.CompareVersions(this.Version, repositoryVersion) switch
+            {
+                > 0 => $"New version: {this.Id} version {this.Version}",
+                < 0 => $"Add version: {this.Id} version {this.Version}",
+                _ => $"Update version: {this.Id} version {this.Version}",
+            };
+        }
```
ryfu-msft, line 758: I feel that this would be better suited in the BaseCommand class as a shared helper function for all commands. Does the submit command also need to call this function to generate its own PR title?

src/WingetCreateCLI/Commands/NewCommand.cs
ryfu-msft, line 255: Since you already have GetPRTitle, can that function also be called here that way the New command is not a special case. This would also make it easier to ensure consistent titles in case we ever need to modify the PR title again in the future.

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/NewCommand.cs
mdanish-kh: New command now calls GetPRTitle() to generate PR title

src/WingetCreateCLI/Commands/UpdateCommand.cs
mdanish-kh: Refactored the method into BaseCommand Class.

mdanish-kh: Currently, the Submit command can't use this function to get its PR title since submit command can be used for any of the conditions (new ver, update, old ver, new pkg). To use this function, Submit command would need to be modified to fetch the manifest from the repo like how Update Command does. The current behavior is that submit command doesn't give any PRTitle info to SubmitGitHubManifest() and the PRtitle for it is set later in GitHub.cs

https://github.com/microsoft/winget-create/blob/aac1dbf8646da27dbb8793bbccecf65d0e89989d/src/WingetCreateCore/Common/GitHub.cs#L307

I'm not sure if we should be fetching an existing manifest in SubmitCommand just to get its PRTitle but I would like some feedback here...

mdanish-kh: Possible option would be either one of:
1. Fetch existing latest manifest from packages repo in SubmitCommand to be sure of type of PR
2. Use a default value in SubmitCommand and don't use optional parameters in SubmitGitHubManifest() (so that we don't need to set it later in GitHub.cs)
3. Leave it as is

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -578,7 +583,40 @@ protected async Task<bool> GitHubSubmitManifests(Manifests manifests, string prT
...
+        protected string GetPRTitle(Manifests currentManifest, Manifests repositoryManifest = null)
+        {
+            // Use custom PR title if provided by the user.
+            if (!string.IsNullOrEmpty(this.PRTitle))
+            {
+                return this.PRTitle;
+            }
+
+            string packageId = currentManifest.SingletonManifest != null ? currentManifest.SingletonManifest.PackageIdentifier : currentManifest.InstallerManifest.PackageIdentifier;
+            string currentVersion = currentManifest.SingletonManifest != null ? currentManifest.SingletonManifest.PackageVersion : currentManifest.InstallerManifest.PackageVersion;
```
ryfu-msft, line 603: Since we don't allow singleton manifests, we should prioritize the version manifest when retrieving these values before resorting to singleton (In theory we should never have to use singleton manifest)

```
string packageId = currentManifest.VersionManifest != null ? currentManifest.Version.PackageIdentifier : currentManifest.SingletonManifest.PackageIdentifier;
```

ryfu-msft: Same for `currentVersion`

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/UpdateCommand.cs
ryfu-msft: Lets go with option 3. I don't want to overcomplicate the submit command by having to fetch existing manifests just to submit a manifest. I think the default PR title in `GitHub.cs` should be kept as a fallback for all submissions so it is okay to have the submit command pass in `null` for the prtitle. Thanks for the thorough explanation.

src/WingetCreateCLI/Commands/BaseCommand.cs
mdanish-kh: Makes sense; Updated to prioritize multi-manifest.

ryfu-msft (approved):

PR #390, opened 2023-06-23
fix: Make sure the enumerable is a List

sandercox:
This allows to setup the `Tags` field which is a List<T>. IEnumerable<T> of the SelectListIterator does not convert to a List<T>. An IEnumerable<T> of List<T> does.

- [x] Are you working against an Issue? - No.

-----

After PR#373 the code was introduced to Trim installer URLs, and trim these properties.

Apparently this works for these URLs that get set to a property of `IEnumerable<string>` but when trying to set the Tags when setting optional default locale fields, this threw an Exception:
```
System.ArgumentException: 'Object of type 'System.Linq.Enumerable+SelectListIterator`2[System.String,System.String]' cannot be converted to type 'System.Collections.Generic.List`1[System.String]'.'
```

Apparently an `IEnumerable<string>` can be set to a `List<string>` when the enumerable comes from an actual `List<string>` but not from a Linq based List.

Turning the Trimmed list back into a real List (it originates from `Prompt.List()`) it seems to work for both URLs and the Tags.
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/390&drop=dogfoodAlpha


ryfu-msft (approved):
Thanks for catching this :)

Issue #397, opened 2023-07-03
ZIP Packages update autonomously but not in interactive mode


mdanish-kh:
### Brief description of your issue

Interactive update uses a call to `ParsePackageAndUpdateInstallerNode()` to Parse the InstallerType of an installer.

https://github.com/microsoft/winget-create/blob/5ec4ad62c6434934c3d630d589bf9c2bf767e264/src/WingetCreateCLI/Commands/UpdateCommand.cs#L724

This uses the Vestris.ResourceLib to parse the installer and retrieve its InstallerType. For zip installers in interactive updates, this seem to throw a Win32 Exception and interactive update returns `Failed to parse the package from [<InstallerUrl>]` at:

https://github.com/microsoft/winget-create/blob/5ec4ad62c6434934c3d630d589bf9c2bf767e264/src/WingetCreateCore/Common/PackageParser.cs#L703-L710

which is called from:
https://github.com/microsoft/winget-create/blob/5ec4ad62c6434934c3d630d589bf9c2bf767e264/src/WingetCreateCore/Common/PackageParser.cs#L283-L285

This is because the `path` passed to the ResourceLib is path to the ZIP archive ending in `.zip` and not the actual installer present in the archive.

### Steps to reproduce

Unattended update: (Dummy version used to simulate update)

1. Nested nullsoft

```powershell
wingetcreate update Guru3D.Afterburner -v 18.0 --urls "https://download-eu2.guru3d.com/afterburner/[Guru3D.com]-MSIAfterburnerSetup465.zip"
```

2. Nested MSI

```powershell
wingetcreate update aluxnimm.outlookcaldavsynchronizer -v 18.0 --urls "https://github.com/aluxnimm/outlookcaldavsynchronizer/releases/download/v4.3.0/OutlookCalDavSynchronizer-4.3.0.zip"
```

3. Nested Portable

```powershell
wingetcreate update Hugo.Hugo -v 18.0 --urls "https://github.com/gohugoio/hugo/releases/download/v0.112.7/hugo_0.112.7_windows-amd64.zip" "https://github.com/gohugoio/hugo/releases/download/v0.112.7/hugo_0.112.7_windows-arm64.zip"
```

4. Nested portable

```powershell
wingetcreate update Istio.Istio -v 18.0 --urls "https://github.com/istio/istio/releases/download/1.17.3/istioctl-1.17.3-win.zip"
```

Notice they all succeed in autonomous update. Now try the same packages in interactive update i.e.,

[...]

Issue #185, opened 2021-10-04
Move as many parameters as possible to Manifest Level


Trenly:
# Description of the new feature/enhancement
Currently the installer type, scope, upgrade behavior, etc. are always placed at the Installer Level. When these values are the same for all installers, they can be moved to the Manifest Level

# Proposed technical implementation details (optional)
Any parameters which are the same across all installers and are valid at both the manifest level and the installer level should be moved to the manifest level during creation or update

jedieaston:
I wonder if this should be enforced in a hypothetical schema 2.0. There aren't a ton of circumstances where putting `InstallerType` or others globally doesn't just turn into a headache later when you want to add a new installer entry.

Trenly:
> I wonder if this should be enforced in a hypothetical schema 2.0. There aren't a ton of circumstances where putting `InstallerType` or others globally doesn't just turn into a headache later when you want to add a new installer entry.

I think that is an interesting point. I think this is mostly a concern when you manually modify the installer entries for new manifests. If you use a tool or a script like the YamlCreate script, then it should handle it automatically. It certainly is something to consider as to whether or not this would be a good thing

denelon:
We've had some discussions around this with the feature team. Some of the optimizations like the singleton format and "root" level nodes that can be overridden in a child node were mostly made for humans crafting manifests. As we're getting closer to solid tooling support, these may be deprecated in favor of a tool generated manifest. It would certainly be a breaking change so the 2.0 schema was a good call out by @jedieaston.

mdanish-kh:
- Resolved in #409

mdanish-kh:
The fix (PR #409) is available now in WinGet-Create 1.5.1.0

cc @denelon / @Trenly if this can be marked as completed

PR #398, opened 2023-07-03
Improve ZIP handling in interactive update

mdanish-kh:
   - #397
   
This change makes it so that the zip archive is extracted in interactive update as well and the path to the nested installer is used in other functions. Made the logic of extracting and determining path of the nested installer into a function as it's common between autonomous and interactive updates.

The logic is based off of the fact that if InstallerType is ZIP, there will always be atleast one NestedInstallerFile node in the manifest that contains RelativeFilePath key and the first matching RelativeFilePath from an older manifest is used as the path to the nested installer. Multiple RelativeFilePath nodes under NestedInstallerFiles are only possible if the NestedInstallerType is portable. In this PR, we do not check the InstallerType of all RelativeFilePath nodes for NestedInstallerType: portable. The assumption is that it's very rare that if an older version contained all portable RelativeFilePath objects, then one of them has changed in newer version to be an another InstallerType (e.g. MSI, nullsoft etc). In case of portables as well, only the first matching RelativeFilePath object is passed to determine the InstallerType to keep the logic simple.

Shifting of values from root to installer is needed in cases where the NestedInstallerFiles object is common between all installer nodes and is present in the root of the manifest instead for better readability. One of the many examples of such manifests is [junnegun.fzf](https://github.com/microsoft/winget-pkgs/blob/master/manifests/j/junegunn/fzf/0.42.0/junegunn.fzf.installer.yaml). Since the interactive update only works by reading the `Installers` node under the installer manifest, we need a way to read the NestedInstallerFiles in the root of the manifest. I did it by shifting all related values to installer level at the start of the interactive update. Ofcourse this comes at a cost of slightly more verbose manifest, but I think it's okay for now considering the benefit of a successful update. We can make another function ShiftInstallerValuesToRoot to take out common fields to the root level but I think that's beyond the scope of this PR as it's applicable to `new`, autonomous `update` commands as well and being tracked in issue https://github.com/microsoft/winget-create/issues/185

   
-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/398&drop=dogfoodAlpha

### DIFF
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -419,6 +419,35 @@ namespace Microsoft.WingetCreateCLI.Commands
         }
 
         /// <summary>
+        /// Shifts common installer fields from root to installer level.
+        /// </summary>
+        /// <param name="installerManifest">Wrapper object containing the installer manifest object models.</param>
+        protected static void ShiftFieldsFromRootToInstallerLevel(InstallerManifest installerManifest)
+        {
+            var rootProperties = installerManifest.GetType().GetProperties();
+            var installerProperties = installerManifest.Installers.First().GetType().GetProperties();
+
+            // Get common properties between root and installer level
+            var commonProperties = rootProperties.Where(p => installerProperties.Any(ip => ip.Name == p.Name));
+
+            foreach (var property in commonProperties)
+            {
+                var rootValue = property.GetValue(installerManifest);
+                if (rootValue != null)
+                {
+                    foreach (var installer in installerManifest.Installers)
+                    {
+                        // Copy the value to installer level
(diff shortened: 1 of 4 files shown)


ryfu-msft (approved):

ryfu-msft:
Apologies for the delay, since the interactive update needs to be manually tested, I needed to checkout your branch and try it manually in order to review this. Shifting the root properties to the installer node is a great addition and it will help improve the interactive flow.

Once thing I wanted to call out was that I found a bug when trying to update junnegun.fzf interactively. All of the architectures get updated to 'x64' thus making the updated manifest invalid. I created an issue here: https://github.com/microsoft/winget-create/issues/405 but I went ahead and approved this PR since that is not related to the feature you just added. Thanks!

PR #409, opened 2023-07-16
Shift common installer fields to manifest root level

mdanish-kh:
  - #185

This change modifies the Update flow (both interactive and autonomous) to shift the manifest root values to installer level at the start of the update for easier program logic and then shifting the common installer values back to manifest root at the end of the update for better human readable manifests. Similar shifting done at the end of New Command flow as well.

Broke `PromptPropertiesAndDisplayManifests` into multiple functions for better separation of concerns and so that shifting can be done before the call to DisplayManifestPreview()

Manually verified with packages like DevHome, Terminal, PowerToys, WinGetCreate to make sure I wasn’t breaking any existing CIs with this change.

Also refactored previous function for consistent naming

-----


src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -445,7 +446,58 @@ protected static void ShiftFieldsFromRootToInstallerLevel(InstallerManifest inst
...
+                {
+                    // Check if all installers have the same value
+                    bool allInstallersHaveSameValue = installerManifest.Installers.All(i =>
+                    {
+                        var propertyValue = i.GetType().GetProperty(property.Name).GetValue(i);
+                        return propertyValue != null &&
+                        propertyValue.Equals(firstInstallerValue);
+                    });
+
+                    // If value is false, it can be becuase we don't have .Equals() override implemented for the type of the property.
```
ryfu-msft, line 476:
```suggestion
                    // If value is false, it can be because we don't have .Equals() override implemented for the type of the property.

```

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -445,7 +446,58 @@ protected static void ShiftFieldsFromRootToInstallerLevel(InstallerManifest inst
...
+                    // For that, we check further if the property is of type list and check if the lists are equal
+                    if (!allInstallersHaveSameValue)
+                    {
+                        if (firstInstallerValue is IList installerValueList)
+                        {
+                            allInstallersHaveSameValue = installerManifest.Installers.All(i =>
+                            i.GetType().GetProperty(property.Name).GetValue(i) is IList otherList &&
+                            installerValueList.Cast<object>().SequenceEqual(otherList.Cast<object>()));
```
ryfu-msft, line 484: Nit: This makes sense but is slightly hard to understand. I think it is better to break up some of the logic here so that it is more readable
```suggestion
                                var propertyValue = i.GetType().GetProperty(property.Name).GetValue(i);
                                if (propertyValue is IList otherList)
                                {
                                    return installerValueList.Cast<object>().SequenceEqual(otherList.Cast<object>());
                                }
                                else
                                {
                                    return false;
                                }
```

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -445,7 +446,58 @@ protected static void ShiftFieldsFromRootToInstallerLevel(InstallerManifest inst
...
+                    });
+
+                    // If value is false, it can be becuase we don't have .Equals() override implemented for the type of the property.
+                    // For that, we check further if the property is of type list and check if the lists are equal
+                    if (!allInstallersHaveSameValue)
+                    {
+                        if (firstInstallerValue is IList installerValueList)
+                        {
+                            allInstallersHaveSameValue = installerManifest.Installers.All(i =>
+                            i.GetType().GetProperty(property.Name).GetValue(i) is IList otherList &&
```
ryfu-msft, line 483: Do we need to check if `i.GetType().GetProperty(property.Name).GetValue(i)` is null before checking if it is of IList type?

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -445,7 +446,58 @@ protected static void ShiftFieldsFromRootToInstallerLevel(InstallerManifest inst
...
                 }
             }
-        }
+        }
+
+        /// <summary>
+        /// Shifts common installer fields from installer level to manifest root.
+        /// </summary>
+        /// <param name="installerManifest">Wrapper object containing the installer manifest object models.</param>
+        protected static void ShiftInstallerFieldsToRootLevel(InstallerManifest installerManifest)
```
ryfu-msft, line 455: Can you add some tests to verify this behavior? I think a good test case would be to update a manifest that has several installers and verifying that all of the common installer fields get shifted over to the root node.

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/BaseCommand.cs
mdanish-kh: if (null is IList installerValueList) will return false and the flow will move to the else statement that'll return false. So, there's no need to explicitly check for null here

src/WingetCreateCLI/Commands/BaseCommand.cs
mdanish-kh: Added a test case to verify the behavior

mdanish-kh:
Took me some time to get around how the tests are added. While I was adding the test case I realized that the properties that are partial classes were not getting compared correctly and we needed to add the .Equals() override for those properties. Also had to redundantly move the methods to autonomous and interactive update methods so that the results are visible in test cases

src/WingetCreateCore/Models/InstallerManifestModels.cs
```diff
@@ -159,8 +160,22 @@ public partial class InstallerSwitches
...
+
+            InstallerSwitches other = (InstallerSwitches)obj;
+            return Silent == other.Silent &&
+                SilentWithProgress == other.SilentWithProgress &&
+                Interactive == other.Interactive &&
+                InstallLocation == other.InstallLocation &&
+                Log == other.Log &&
+                Upgrade == other.Upgrade &&
+                Custom == other.Custom;
+        }
```
ryfu-msft, line 178: The InstallerManifestModel.cs file cannot be modified as it is autogenerated based on the latest schema. 

You will need to move these functions to `WingetCreateCore/Models/InstallerPartials` and follow the pattern there. Make sure you do that for the other `Equals` methods so that this file has no changes and remove the `using System.Linq` statement.

src/WingetCreateTests/WingetCreateTests/WingetCreateTests.csproj
```diff
@@ -108,6 +111,9 @@
     <None Update="Resources\TestPublisher.MismatchedMsixInExistingBundle.yaml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+	<None Update="Resources\TestPublisher.MoveInstallerFieldsToRoot.yaml">
+	  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+	</None>
```
ryfu-msft, line 116: nit: alignment

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -829,6 +823,152 @@ public async Task UpdateZipWithMsix()
...
+            Assert.IsNotNull(updatedInstallerManifest.InstallerSwitches, "InstallerSwitches at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.Dependencies, "Dependencies at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.AppsAndFeaturesEntries, "AppsAndFeaturesEntries at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.Platform, "Platform at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.ExpectedReturnCodes, "ExpectedReturnCodes at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.Commands, "Commands at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.Protocols, "Protocols at the root level should not be null");
+            Assert.IsNotNull(updatedInstallerManifest.FileExtensions, "FileExtensions at the root level should not be null");
+            // TODO: Uncomment when installer model gets updated to support markets field.
```
ryfu-msft, line 861: Nit: There should be an empty line preceding this comment. Same for other TODO comment below.

ryfu-msft (changes requested):
Nice tests! Just a couple more modifications and this should be good to go.

src/WingetCreateTests/WingetCreateTests/WingetCreateTests.csproj
mdanish-kh: Apparently, VS wasn't showing me this alignment difference (I'm sure of my eyes 😄). Deleted and re-added the file so that the line was auto-generated. Confirmed that the alignment is correct in another editor

src/WingetCreateCore/Models/InstallerManifestModels.cs
mdanish-kh: Shifted the methods to `WingetCreateCore/Models/InstallerPartials` and reset changes to `InstallerManifestModel.cs`

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
mdanish-kh: Fixed in the latest commit

mdanish-kh:
Let me know if I messed up again while addressing the review comments 😄 I would want to make it right before we accept this PR. VS is warning me to implement GetHashCode() overrides and separate each partial class into a separate file. Let me know if this is something I should look into or something that we can ignore for now.

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+            PackageDependencies other = (PackageDependencies)obj;
+            return this.PackageIdentifier == other.PackageIdentifier &&
+                this.MinimumVersion == other.MinimumVersion;
         }
     }
 
+    /// <summary>
+    /// Partial class that implements helper methods for the InstallerSwitches class.
+    /// </summary>
+    public partial class InstallerSwitches
```
ryfu-msft, line 49:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class InstallerSwitches
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+                this.Log == other.Log &&
+                this.Upgrade == other.Upgrade &&
+                this.Custom == other.Custom;
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the Dependencies class.
+    /// </summary>
+    public partial class Dependencies
```
ryfu-msft, line 77:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class Dependencies
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+                this.WindowsLibraries.ToList().SequenceEqual(this.WindowsLibraries) &&
+                this.PackageDependencies.ToList().SequenceEqual(other.PackageDependencies) &&
+                this.ExternalDependencies.ToList().SequenceEqual(other.ExternalDependencies);
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the AppsAndFeaturesEntry class.
+    /// </summary>
+    public partial class AppsAndFeaturesEntry
```
ryfu-msft, line 102:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class AppsAndFeaturesEntry
#pragma warning restore SA1402 // File may only contain a single type
```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+                this.ProductCode == other.ProductCode &&
+                this.UpgradeCode == other.UpgradeCode &&
+                this.InstallerType == other.InstallerType;
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the InstallationMetadata class.
+    /// </summary>
+    public partial class InstallationMetadata
```
ryfu-msft, line 129:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class InstallationMetadata
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+            InstallationMetadata other = (InstallationMetadata)obj;
+            return this.DefaultInstallLocation == other.DefaultInstallLocation &&
+                this.Files.ToList().SequenceEqual(other.Files);
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the Files class.
+    /// </summary>
+    public partial class Files
```
ryfu-msft, line 152:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class Files
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+                this.FileType == other.FileType &&
+                this.InvocationParameter == other.InvocationParameter &&
+                this.DisplayName == other.DisplayName;
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the NestedInstallerFile class.
+    /// </summary>
+    public partial class NestedInstallerFile
```
ryfu-msft, line 178:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class NestedInstallerFile
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+            NestedInstallerFile other = (NestedInstallerFile)obj;
+            return this.RelativeFilePath == other.RelativeFilePath &&
+                this.PortableCommandAlias == other.PortableCommandAlias;
+        }
+    }
+
+    /// <summary>
+    /// Partial class that implements helper methods for the ExpectedReturnCode class.
+    /// </summary>
+    public partial class ExpectedReturnCode
```
ryfu-msft, line 201:
```suggestion
#pragma warning disable SA1402 // File may only contain a single type
    public partial class ExpectedReturnCode
#pragma warning restore SA1402 // File may only contain a single type

```

src/WingetCreateCore/Models/Partials/InstallerPartials.cs
```diff
@@ -22,9 +23,202 @@ public partial class PackageDependencies : ICloneable
...
+    {
+        /// <summary>
+        /// Gives the criteria for determining whether two instances of ExpectedReturnCode objects are equal
```
ryfu-msft, line 204:
```suggestion
        /// Gives the criteria for determining whether two instances of ExpectedReturnCode objects are equal.

```

ryfu-msft (changes requested):
We want to make sure these [build warnings](https://dev.azure.com/ms/winget-create/_build/results?buildId=477772&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=1b55af13-a9e6-55b7-5bf5-86fbedb24a38&l=52#:~:text=C%3A%5CProgram%20Files,with%20a%20period) go away as I think these may cause issues when we try to publish a new release. Don't worry about the GetHashCode() warnings, it is fine as is.

ryfu-msft (approved):
:shipit:

Issue #414, opened 2023-07-26
wingetcreate update fails while there is a version number in the zip archive


lonnywong:
### Brief description of your issue

The zip archive file `tssh_0.1.8_windows_x86_64.zip` having `tssh_0.1.8_windows_x86_64\tssh.exe` in it. 

The version number `0.1.8` causes `wingetcreate update` to fail.

### Steps to reproduce

```
wingetcreate update --urls `
https://github.com/trzsz/trzsz-ssh/releases/download/v0.1.8/tssh_0.1.8_windows_x86_64.zip `
https://github.com/trzsz/trzsz-ssh/releases/download/v0.1.8/tssh_0.1.8_windows_aarch64.zip `
https://github.com/trzsz/trzsz-ssh/releases/download/v0.1.8/tssh_0.1.8_windows_i386.zip `
--version 0.1.8 trzsz.tssh
```

### Expected behavior

wingetcreate update success.

### Actual behavior

```
Retrieving latest manifest for trzsz.tssh
Downloading and parsing: https://github.com/trzsz/trzsz-ssh/releases/download/v0.1.8/tssh_0.1.8_windows_x86_64.zip...
Nested installer not found in zip archive: tssh_0.1.7_windows_x86_64\tssh.exe
````

### Environment

```shell
Windows Package Manager Manifest Creator v1.2.8.0
```

```
Windows Package Manager v1.5.1881
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.1992
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.20.1881.0

Winget Directories
-------------------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\DiagOutputDir
User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\settings.json
Portable Links Directory (User)    %LOCALAPPDATA%\Microsoft\WinGet\Links
Portable Links Directory (Machine) C:\Program Files\WinGet\Links
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portable Package Root              C:\Program Files\WinGet\Packages
Portable Package Root (x86)        C:\Program Files (x86)\WinGet\Packages

Links
---------------------------------------------------------------------------
[...]

Issue #391, opened 2023-06-23
CRLF line endings for all sources files


sandercox:
### Brief description of your issue

Trying to modify some code in the repo and doing a PR was almost impossible to get just my change to show. The file I tried to edit had mixed line endings.

### Steps to reproduce

Try and open `src/WingetCreateCLI/PromptHelper.cs` with VS code and add /remove a line, make a commit and see that hundreds of lines have changed.

### Expected behavior

After making a change and saving according to editorconfig settings, only my change is in the diff.

Maybe add a github action to have a double check on this policy such that no bad line endings end up in the source afterwards.

### Actual behavior

All lines are reformatted in a decent editor creating messy PRs

### Environment

```shell
-
```

PR #415, opened 2023-07-26
Get new RelativeFilePaths if previous do not exist

mdanish-kh:
  - #414

This change makes it so we take the last portion of the RelativeFilePath i.e., the file name and search that file name in the new archive for a match. If it is a unique match, we get its RelativeFilePath and use it for the update.

Known limitation is that there are some cases where there may be more than one match in the archive (An example of this being [DynamoRIO.DynamoRIO package](https://github.com/microsoft/winget-pkgs/blob/master/manifests/d/DynamoRIO/DynamoRIO/9.93.19518/DynamoRIO.DynamoRIO.installer.yaml)). This is not handled in this PR and we print out an appropriate message to the user. I'll create an issue for it if this PR gets accepted.

The change to use the whole NestedInstallerFile object instead of previous RelativeFilePaths is so that I can replace this whole object in UpdateInstallerMetadata() and have it updated.

### Test packages

Somes packages I found to test this PR with

```powershell
wingetcreate update Smallstep.step -v 18.0 --urls "https://github.com/smallstep/cli/releases/download/v0.24.4/step_windows_0.24.4_amd64.zip" "https://github.com/smallstep/cli/releases/download/v0.24.4/step_windows_0.24.4_arm64.zip"
```

```powershell
wingetcreate update Genymobile.scrcpy -v 18.0 --urls "https://github.com/Genymobile/scrcpy/releases/download/v1.25/scrcpy-win64-v1.25.zip" "https://github.com/Genymobile/scrcpy/releases/download/v1.25/scrcpy-win32-v1.25.zip"
```

Multiple matches for RelativeFilePaths (**not handled in this PR**)
```powershell
wingetcreate update DynamoRIO.DynamoRIO --version 18.0 --urls "https://github.com/DynamoRIO/dynamorio/releases/download/cronbuild-9.93.19557/DynamoRIO-Windows-9.93.19557.zip"
```

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/415&drop=dogfoodAlpha

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -291,15 +291,15 @@ namespace Microsoft.WingetCreateCLI.Commands
                         if (!File.Exists(Path.Combine(extractDirectory, nestedInstallerFile.RelativeFilePath)))
                         {
                             string relativeFilePath = ObtainMatchingRelativeFilePath(nestedInstallerFile.RelativeFilePath, extractDirectory, packageFile);
-                            if (relativeFilePath == null)
+                            if (string.IsNullOrEmpty(relativeFilePath))
                             {
                                 return null;
                             }
 
-                            List<string> relativeFilePathsInsideList = installerUpdate.NestedInstallerFiles.Select(x => x.RelativeFilePath).ToList();
+                            List<string> installerUpdateRelativeFilePathList = installerUpdate.NestedInstallerFiles.Select(x => x.RelativeFilePath).ToList();
 
-                            // Skip if the relative path is already in the list.
-                            if (!relativeFilePathsInsideList.Contains(relativeFilePath))
+                            // Skip if the relative path is already in the installer update list.
+                            if (!installerUpdateRelativeFilePathList.Contains(relativeFilePath))
                             {
                                 installerUpdate.NestedInstallerFiles.Add(new NestedInstallerFile
                                 {
@@ -448,7 +448,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             // If there is only one match in the archive, use that as the new relative file path.
             if (matchingFiles.Length == 1)
             {
-                string relativePath = matchingFiles.First().Replace(directory, string.Empty).TrimStart('\\');
+                string relativePath = matchingFiles.First().Replace(directory, string.Empty).TrimStart(Path.DirectorySeparatorChar);
                 return relativePath;
             }
             else if (matchingFiles.Length > 1)
@@ -802,7 +802,7 @@ namespace Microsoft.WingetCreateCLI.Commands
                         foreach (NestedInstallerFile nestedInstaller in installer.NestedInstallerFiles)
                         {
                             nestedInstaller.RelativeFilePath = ObtainMatchingRelativeFilePath(nestedInstaller.RelativeFilePath, extractDirectory, packageFile);
-                            if (nestedInstaller.RelativeFilePath == null)
+                            if (string.IsNullOrEmpty(nestedInstaller.RelativeFilePath))
                             {
                                 continue;
                             }
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -290,34 +290,22 @@ namespace Microsoft.WingetCreateCLI.Commands
                         // If the previous RelativeFilePath is not an exact match, check if there is a new suitable match in the archive.
                         if (!File.Exists(Path.Combine(extractDirectory, nestedInstallerFile.RelativeFilePath)))
                         {
-                            string fileName = Path.GetFileName(nestedInstallerFile.RelativeFilePath);
-                            string[] matchingFiles = Directory.GetFiles(extractDirectory, fileName, SearchOption.AllDirectories);
-
-                            // If there is only one match in the archive, use that as the new relative file path.
-                            if (matchingFiles.Length == 1)
+                            string relativeFilePath = ObtainMatchingRelativeFilePath(nestedInstallerFile.RelativeFilePath, extractDirectory, packageFile);
+                            if (relativeFilePath == null)
                             {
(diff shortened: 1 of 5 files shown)


mdanish-kh:
The GitHub diffs are not correct after https://github.com/microsoft/winget-create/pull/415/commits/83973aa632218217c09ea7a330898a3a8e0083c0. I only modified the UpdateSingleInstallerInteractively method but that method had both mixed line endings (CRLF and LF) because of which the diffs are messed up. If you check out the changes in Visual Studio, it'll show you the correct diffs. 
- Related to https://github.com/microsoft/winget-create/issues/391

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -629,124 +673,140 @@ private List<InstallerMetadata> ParseInstallerUrlsForOverrides(List<string> inst
...
-                    else
-                    {
-                        Logger.ErrorLocalized(nameof(Resources.UnmatchedNestedInstaller_Error), relativeFilePath, url);
-                        continue;
+
+                        // Update the relative file path for all nested installers.
+                        foreach (NestedInstallerFile nestedInstaller in installer.NestedInstallerFiles)
+                        {
+                            nestedInstaller.RelativeFilePath = ObtainMatchingRelativeFilePath(nestedInstaller.RelativeFilePath, extractDirectory, packageFile);
+                            if (nestedInstaller.RelativeFilePath == null)
```
ryfu-msft, line 805: nit:
```suggestion
                            if (string.IsNullOrEmpty(nestedInstaller.RelativeFilePath))

```

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -419,6 +440,29 @@ private static string ExtractArchiveAndRetrieveDirectoryPath(string packageFileP
...
 
+        private static string ObtainMatchingRelativeFilePath(string oldRelativeFilePath, string directory, string archiveName)
+        {
+            string fileName = Path.GetFileName(oldRelativeFilePath);
+            string[] matchingFiles = Directory.GetFiles(directory, fileName, SearchOption.AllDirectories);
+
+            // If there is only one match in the archive, use that as the new relative file path.
+            if (matchingFiles.Length == 1)
+            {
+                string relativePath = matchingFiles.First().Replace(directory, string.Empty).TrimStart('\\');
```
ryfu-msft, line 451:
```suggestion
                string relativePath = matchingFiles.First().Replace(directory, string.Empty).TrimStart(Path.DirectorySeparatorChar);

```

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -280,21 +280,42 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
...
                     {
-                        if (!File.Exists(Path.Combine(extractDirectory, relativeFilePath)))
+                        // If the previous RelativeFilePath is not an exact match, check if there is a new suitable match in the archive.
+                        if (!File.Exists(Path.Combine(extractDirectory, nestedInstallerFile.RelativeFilePath)))
                         {
-                            Logger.ErrorLocalized(nameof(Resources.NestedInstallerFileNotFound_Error), relativeFilePath);
-                            return null;
-                        }
+                            string relativeFilePath = ObtainMatchingRelativeFilePath(nestedInstallerFile.RelativeFilePath, extractDirectory, packageFile);
+                            if (relativeFilePath == null)
```
ryfu-msft, line 294: nit:
```suggestion
                if (string.IsNullOrEmpty(packageFile))
```

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -280,21 +280,42 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
...
+                            {
+                                return null;
+                            }
+
+                            List<string> relativeFilePathsInsideList = installerUpdate.NestedInstallerFiles.Select(x => x.RelativeFilePath).ToList();
```
ryfu-msft, line 299: Nit: This variable name was slightly confusing for me. I think something like `installerUpdateRelativeFilePathList` might be better.

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -280,21 +280,42 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
...
 
-                        installerUpdate.RelativeFilePaths.Add(relativeFilePath);
+                            // Skip if the relative path is already in the list.
```
ryfu-msft, line 301:
```suggestion
                            // Skip if the relative path is already in the installer update list.

```

ryfu-msft (changes requested):
Just a couple nit comments

ryfu-msft:
> The GitHub diffs are not correct after [83973aa](https://github.com/microsoft/winget-create/commit/83973aa632218217c09ea7a330898a3a8e0083c0). I modified the UpdateSingleInstallerInteractively method but that method had both mixed line endings (CRLF and LF) because of which the diffs are messed up. If you check out the changes in Visual Studio, it'll show you the correct diffs.
> 
> * Related to [CRLF line endings for all sources files #391](https://github.com/microsoft/winget-create/issues/391)
> 
> If you like I can fix the line endings for UpdateCommand.cs and PackageParser.cs in a new commit as they are the two most high traffic files, though that'll mess the diffs up even more for this PR

If you can do this in a separate PR (after this gets checked in) that would be a huge help! 🙏 
That way we can link the related issue to that PR when it gets checked in.

mdanish-kh:
> If you can do this in a separate PR (after this gets checked in) that would be a huge help! 🙏 That way we can link the related issue to that PR when it gets checked in.

Gotcha! Will do after this one. I was shying away from fixing the line endings since it'll mess up the `git blame` output but I guess fixing it has more value.

ryfu-msft (approved):

PR #416, opened 2023-07-30
Configuration file for WinGet-Create

mdanish-kh:
   No; In spirit of adding configuration files to projects, made one for WinGet-Create.

.vsconfig is generated by selecting `.NET Desktop Development` and `Universal Windows Platform Development` workloads including their optional components that get auto-selected when you pick these in VS Installer. Tested the successful configuration in a clean VM.

Note: winget-create doesn’t strictly require Developer mode to build the client, but VS throws a pop-up warning if it isn’t enabled so I decided to include it anyway.

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/416&drop=dogfoodAlpha


.configurations/configuration.dsc.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - resource: Microsoft.VisualStudio.DSC/VSComponents
+      dependsOn:
+        - vsPackage
+      directives:
+        description: Install required VS workloads
+        allowPrerelease: true
+      settings:
+        productId: Microsoft.VisualStudio.Product.Community
+        channelId: VisualStudio.17.Release
+        vsConfigFile: '${WinGetConfigRoot}\..\.vsconfig'
```
ryfu-msft, line 28: There is another parameter you can use here to include recommended and optional components: Link for your reference: https://github.com/microsoft/VisualStudioDSC/blob/582d5fead3fcffb2796a193c7f357973798b7df4/Microsoft.VisualStudio.DSC/Microsoft.VisualStudio.DSC.psm1#L24-L29

I think you should only add `includedRecommended = true` here. We don't want to force users to include optional components if they decide to use this configuration file. It would be better to configure what is needed and let the user add the optional components as they see fit.

.vsconfig
```diff
@@ -0,0 +1,47 @@
+{
+  "version": "1.0",
+  "components": [
```
ryfu-msft, line 3: As mentioned in the previous comment, we should just included what's needed but set `includeRecommended = true`  in the settings:
```
          "Microsoft.VisualStudio.Workload.ManagedDesktop"
          "Microsoft.VisualStudio.Workload.Universal"
```

ryfu-msft (changes requested):

mdanish-kh:
Thanks! Didn't know about `includeRecommended: true`. Verified the new configuration in a clean VM.

PR #427, opened 2023-08-11
Add support for removing temporary files

### LINKED ISSUE #426: Clean up log files and installer cache directory

If a person uses winget-create regularly, the installer cache directory can hog up a lot of space without the user realizing it. I usually use the `wingetcreate cache --clean` command but for a few weeks when I didn't while testing PRs I realized the directory took more than 8 gigs of space

Similar to how winget-cli removes log files:
- https://github.com/microsoft/winget-cli/issues/3434

winget-create should also periodically clean up installer cache and logs directory

mdanish-kh:
	- Resolves #426

Also added a user setting to control the cleanup process. Chose 7-day interval as the default to be consistent with https://github.com/microsoft/winget-cli/issues/3434

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/427&drop=dogfoodAlpha

### DIFF
--- a/doc/settings.md
+++ b/doc/settings.md
@@ -21,7 +21,7 @@ If set to true, the `telemetry.disable` setting will prevent any event from bein
 
 ## CleanUp
 
-The `CleanUp`` settings determine whether Winget-Create will handle the removal of temporary files (installer cache and logs) generated during the manifest creation process. These settings provide control over the decision to remove files or not and the frequency at which this clean up occurs.
+The `CleanUp` settings determine whether Winget-Create will handle the removal of temporary files (installer cache and logs) generated during the manifest creation process. These settings provide control over the decision to remove files or not and the frequency at which this clean up occurs.
 
 ### disable
 
--- a/doc/settings.md
+++ b/doc/settings.md
@@ -19,6 +19,30 @@ See [details on telemetry](https://github.com/microsoft/winget-create#datateleme
 
 If set to true, the `telemetry.disable` setting will prevent any event from being written by the program.
 
+## CleanUp
+
+The `CleanUp`` settings determine whether Winget-Create will handle the removal of temporary files (installer cache and logs) generated during the manifest creation process. These settings provide control over the decision to remove files or not and the frequency at which this clean up occurs.
+
+### disable
+
+```json
+    "CleanUp": {
+        "disable": true
+    },
+```
+
+If set to true, the `CleanUp.disable` setting will prevent any temporary files from being removed by the program.
+
+### intervalInDays
+
+```json
+    "CleanUp": {
+        "intervalInDays": 7
+    },
+```
+
+The `intervalInDays` setting specifies how often Winget-Create will remove temporary files. By default, this is set to 7 days.
+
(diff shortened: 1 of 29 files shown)


doc/settings.md
```diff
@@ -19,6 +19,30 @@ See [details on telemetry](https://github.com/microsoft/winget-create#datateleme
...
 
+## CleanUp
+
+The `CleanUp`` settings determine whether Winget-Create will handle the removal of temporary files (installer cache and logs) generated during the manifest creation process. These settings provide control over the decision to remove files or not and the frequency at which this clean up occurs.
+
+### disable
+
+```json
+    "CleanUp": {
+        "disable": true
```
Trenly, line 30: Is this actually needed? Couldn't a user just set the intervalInDays to `-1` to effectively disable the cleanup?

mdanish-kh: Had that in mind but decided against it since I thought that wouldn't immediately be obvious to the user opening the settings in VSCode and would probably require referring to the documentation. I'm fine either way, I'll wait if @ryfu-msft agrees with the suggestion and I'll update the schema

Trenly: I was just curious more than anything. Doesn't really make a difference either way

ryfu-msft: I think having a boolean that explicitly enables/disables this feature is probably the safest route. However, I do think non-default behavior should always be set to false. 

In other words, I think this should be changed to `CleanUp.enable = false` as the default case.

src/WingetCreateCLI/Common.cs
```diff
@@ -28,6 +28,37 @@ public static class Common
         /// </summary>
         public static string LocalAppStatePath => AppStatePathLazy.Value;
 
+        /// <summary>
+        /// Initiates the clean up process for the given directory.
+        /// </summary>
+        /// <param name="cleanUpDirectory">Directory to clean up.</param>
+        public static void BeginCleanUp(string cleanUpDirectory)
```
ryfu-msft, line 35: Since this is a common function, it should be written in a way that could be used in other scenarios (Even if at this time it only applies to the cleanup settings). 

I would pass in the cleanupDays as a parameter instead of getting that value directly from settings within this method. Also rename the method so that it explicitly tells us what it is doing like `DeleteFilesOlderThan` (not the best name but just an example)

ryfu-msft (changes requested):
Super neat addition, just a few comments :)

doc/settings.md
ryfu-msft: That's a good point. I am okay with making it a default behavior as I'd rather have proper cleanup to save space for our users so they're not storing a bunch of installers.

src/WingetCreateCLI/Common.cs
mdanish-kh: Honestly couldn't think of a better name as well 😄 I went with `CleanUpFilesOlderThan`, made the function suitable for calling in other scenarios as well

ryfu-msft (approved):

PR #433, opened 2023-08-14
Bump YamlDotNet to 13.2.0

### LINKED ISSUE #417: When Version is a Floating Point Number, it is Not Quoted

If package version is a floating point number, it doesn't get quoted
https://repology.org/log/1208825

### Steps to reproduce

Use a package version like 6.81

### Expected behavior

PackageVersion: "6.81"

### Actual behavior

PackageVersion: 6.81

```
# Created using wingetcreate 1.2.7.0
# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.4.0.schema.json

PackageIdentifier: Cockos.REAPER
PackageVersion: 6.81
Platform:
- Windows.Desktop
```

### Environment

```shell
.
```

mdanish-kh:
	- Resolves #417
	
This change now wraps all string values in the manifest around ~double quotes~ single quotes, we can choose from either but I went with one that looks cleaner. Single quotes don't parse escape sequences (not that it matters for our use case i think). (though imo this change makes the overall manifest look ugly now with everything quoted but that's just personal preference 😄)

Tested manually with

```powershell
wingetcreate update Cockos.REAPER --urls "https://www.reaper.fm/files/6.x/reaper681_x64-install.exe|x64" "https://www.reaper.fm/files/6.x/reaper681-install.exe|x86"
```

```powershell
wingetcreate update OpenAL.OpenAL --urls "https://www.openal.org/downloads/oalinst.zip"
```

-----

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/winget-create/pull/433&drop=dogfoodAlpha

### DIFF
--- a/src/WingetCreateCore/Common/Serialization.cs
+++ b/src/WingetCreateCore/Common/Serialization.cs
@@ -42,7 +42,7 @@ namespace Microsoft.WingetCreateCore
         {
             var serializer = new SerializerBuilder()
                 .WithNamingConvention(PascalCaseNamingConvention.Instance)
-                .WithDefaultScalarStyle(ScalarStyle.DoubleQuoted)
+                .WithDefaultScalarStyle(ScalarStyle.SingleQuoted)
                 .WithTypeConverter(new YamlStringEnumConverter())
                 .WithEmissionPhaseObjectGraphVisitor(args => new YamlSkipPropertyVisitor(args.InnerVisitor))
                 .WithEventEmitter(nextEmitter => new MultilineScalarFlowStyleEmitter(nextEmitter))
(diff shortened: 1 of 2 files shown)


src/WingetCreateCore/Common/Serialization.cs
```diff
@@ -42,6 +42,7 @@ public static ISerializer CreateSerializer()
         {
             var serializer = new SerializerBuilder()
                 .WithNamingConvention(PascalCaseNamingConvention.Instance)
+                .WithDefaultScalarStyle(ScalarStyle.SingleQuoted)
```
Trenly, line 45:
```suggestion
               .WithQuotingNecessaryStrings()
```

Trenly (commented):
Found it (suggested above)

image

Issue #72, opened 2021-06-09
`wingetcreate update` if latest InstallerUrl


OfficialEsco:
# Description of the new feature/enhancement
`wingetcreate update` should delete the old version folder if it uses the same `InstallerUrl` and count as `Update and Rename` so we or the bot don't have to clean up the folder afterwards.

denelon:
@OfficialEsco do you have an example PR showing what this looks like? I did a quick search and didn't find an example.

OfficialEsco:
That's because you're looking too far https://github.com/microsoft/winget-pkgs/pull/16874/files
This is how it SHOULD look, if you move a few lines it will count as a new file. Wingetbot/WinGetcreate is therefore incompetable with each other.

palenshus:
The latter case is probably due to line endings changing. Anyway, I'm wondering if in cases like this the version shouldn't just be changed to "latest" in the manifest, and that way the directory doesn't change from update to update?

OfficialEsco:
Well wingetbot only updates the hash, I can do a lot bigger changes without the whole file getting recreated...

We have stopped using Latest PackageVersion as it causes issues with winget upgrade.

PR #440, opened 2023-08-27
Replace argument for removing a previous manifest

### LINKED ISSUE #277: Add a "replace" argument to `update`

Some packages only publish a single version at a time. As a contributor, I'd like to be able to use wingetcreate to create a new version, and in the same commit, remove the old version so that I don't have to follow up in a second PR to remove the currently broken version. 

A `--replace` argument for `wingetcreate update`, optionally taking a version number (but if none is given and there's a single version currently in the community repo, then it should just know to remove the only existing one).

mdanish-kh:
	- Resolves #277
	
Example PR made with `--replace`: 
- https://github.com/microsoft/winget-pkgs/pull/118148

There doesn't seem to be a _great_ way to make an option accept optional values. Followed the recommendations listed here: https://github.com/commandlineparser/commandline/issues/674

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/440)

### DIFF
--- a/doc/update.md
+++ b/doc/update.md
@@ -54,7 +54,7 @@ The following arguments are available:
 | **-v, --version** |  Version to be used when updating the package version field.
 | **-o, --out** |  The output directory where the newly created manifests will be saved locally
 | **-s, --submit** |  Boolean value for submitting to the Windows Package Manager repo. If true, updated manifest will be submitted directly using the provided GitHub Token
-| **-r, --replace** |  Boolean value for removing an existing manifest from the Windows Package Manager repo. Optionally provide a version or else the latest version will be removed. Default is false.
+| **-r, --replace** |  Boolean value for replacing an existing manifest from the Windows Package Manager repo. Optionally provide a version or else the latest version will be replaced. Default is false.
 | **-p, --prtitle** |  The title of the pull request submitted to GitHub.
 | **-t, --token** |  GitHub personal access token used for direct submission to the Windows Package Manager repo. If no token is provided, tool will prompt for GitHub login credentials.
 | **-?, --help** |  Gets additional help on this command. |
--- a/doc/update.md
+++ b/doc/update.md
@@ -1,24 +1,25 @@
 
 # update command (Winget-Create)
 
-The **update** command of the [Winget-Create](../README.md) tool is designed to update an existing manifest. The **update** command is non-interactive so that it can be seamlessly integrated into your build pipeline to assist with the publishing of your installer.  The **update** command will update the manifest with the new URL, hash and version and can automatically submit the pull request (PR) to the [Windows Package Manager repo](https://docs.microsoft.com/windows/package-manager/).  
+The **update** command of the [Winget-Create](../README.md) tool is designed to update an existing manifest. The **update** command is non-interactive so that it can be seamlessly integrated into your build pipeline to assist with the publishing of your installer.  The **update** command will update the manifest with the new URL, hash and version and can automatically submit the pull request (PR) to the [Windows Package Manager repo](https://docs.microsoft.com/windows/package-manager/).
 
 ## Usage
 
-`wingetcreate.exe update <id> [-u <urls>] [-v <version>] [-s] [-t <token>] [-o <output directory>]`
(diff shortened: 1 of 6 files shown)


mdanish-kh:
I will be following up with a PR to have "replace" work automatically for vanity URLs provided this gets approved

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -144,6 +156,26 @@ public override async Task<bool> Execute()
                     this.Id = exactId;
                 }
 
+                if (!string.IsNullOrEmpty(this.ReplaceVersion))
+                {
+                    // If update version is same as replace version, it's a regular update.
+                    if (this.Version == this.ReplaceVersion)
+                    {
+                        this.Replace = false;
+                    }
```
ryfu-msft, line 165: I think it would be better to show an error message to the user that the replacement version matches the update version since that should not be a supported scenario. That way if the user accidentally types in an identical replacement version, they are notified rather than performing an update.

doc/update.md
```diff
@@ -53,14 +54,17 @@ The following arguments are available:
 | **-v, --version** |  Version to be used when updating the package version field.
 | **-o, --out** |  The output directory where the newly created manifests will be saved locally
 | **-s, --submit** |  Boolean value for submitting to the Windows Package Manager repo. If true, updated manifest will be submitted directly using the provided GitHub Token
+| **-r, --replace** |  Boolean value for removing an existing manifest from the Windows Package Manager repo. Optionally provide a version or else the latest version will be removed. Default is false.
```
ryfu-msft, line 57: Change all other strings that use this description to use `replace` rather than `remove`
```suggestion
| **-r, --replace** |  Boolean value for replacing an existing manifest from the Windows Package Manager repo. Optionally provide a version or else the latest version will be replaced. Default is false.
```

src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -406,6 +396,53 @@ private async Task<PullRequest> SubmitPRAsync(string packageId, string version,
...
+            if (string.IsNullOrEmpty(versionDirectoryPath))
+            {
+                throw new NotFoundException(nameof(version), System.Net.HttpStatusCode.NotFound);
+            }
+
+            // Get all files in the version directory
+            var versionDirectoryContents = await this.github.Repository.Content.GetAllContents(forkRepoId, versionDirectoryPath);
+
+            // Delete files from the new branch in the forked repository
+            foreach (var file in versionDirectoryContents)
```
ryfu-msft, line 439: Do we also need to handle cleaning up the folder of the old version manifests?

ryfu-msft (changes requested):

src/WingetCreateCore/Common/GitHub.cs
mdanish-kh: No we don't. That is handled automatically by GitHub/Git (as git can track only "files" by default). To verify, you can see this in action with a test PR i made:

https://github.com/mdanish-kh/winget-pkgs/pull/65

master branch: https://github.com/mdanish-kh/winget-pkgs/tree/master/manifests/a/Anydo/Anydo
PR branch: https://github.com/mdanish-kh/winget-pkgs/tree/Anydo.Anydo-5.0.40-121efaa7-325a-4275-9680-20f8ae006913/manifests/a/Anydo/Anydo

The folder `5.0.38.17924` was automatically removed.

ryfu-msft:
@mdanish-kh , Once this PR gets checked, I will work on pulling in the new localized strings from this PR once they are generated and cutting a new release :)

mdanish-kh:
@ryfu-msft Thanks, I do have a quick PR incoming after this (to address #72) but I don't mind if that can wait until the next release

ryfu-msft:
> @ryfu-msft Thanks, I do have a quick PR incoming after this (to address #72) but I don't mind if that can wait until the next release

I'm happy to wait for that since that will be a really nice addition

ryfu-msft (approved):

PR #446, opened 2023-09-10
Recommend using overrides in error message

mdanish-kh:
	No, this is to improve error messages to redirect user to a possible fix

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/446)

### DIFF
--- a/src/WingetCreateCLI/Commands/UpdateCommand.cs
+++ b/src/WingetCreateCLI/Commands/UpdateCommand.cs
@@ -351,6 +351,8 @@ namespace Microsoft.WingetCreateCLI.Commands
                 Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
                 installerMatchException.MultipleMatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.UnmatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
                 installerMatchException.UnmatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.MultipleMatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
+                Console.WriteLine();
+                Logger.ErrorLocalized(nameof(Resources.UseOverrides_ErrorMessage));
 
                 if (installerMatchException.IsArchitectureOverride)
                 {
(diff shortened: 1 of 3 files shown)


src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -351,6 +351,8 @@ public async Task<Manifests> UpdateManifestsAutonomously(Manifests manifests)
                 Logger.ErrorLocalized(nameof(Resources.NewInstallerUrlMustMatchExisting_Message));
                 installerMatchException.MultipleMatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.UnmatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
                 installerMatchException.UnmatchedInstallers.ForEach(i => Logger.ErrorLocalized(nameof(Resources.MultipleMatchedInstaller_Error), i.Architecture, i.InstallerType, i.InstallerUrl));
+                Console.WriteLine();
+                Logger.ErrorLocalized(nameof(Resources.UseOverrides_ErrorMessage));
 
                 if (installerMatchException.IsArchitectureOverride)
```
ryfu-msft, line 357: The placement here will suggest to the user that they should use an override if the installers fail to match. The if statement below that will also show a warning to the user if the architecture override does not work which is redundant. 

I would change this to `Logger.WarnLocalized` and put it in an else clause so that it only warns the user if they do not provide an architecture override.

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1099,4 +1099,7 @@
   <data name="VersionManifest_Message" xml:space="preserve">
     <value>Version Manifest:</value>
   </data>
+  <data name="UseOverrides_ErrorMessage" xml:space="preserve">
+    <value>Please try using the architecture and/or scope overrides to uniquely match new URLs to existing installer nodes in the manifest.</value>
```
ryfu-msft, line 1103: 'Please' should not be used in warning or error messages
```suggestion
    <value>Try using the architecture and/or scope overrides to uniquely match new URLs to existing installer nodes in the manifest.</value>

```

ryfu-msft (changes requested):

src/WingetCreateCLI/Properties/Resources.resx
mdanish-kh: Thanks, I'll keep that in mind! Resolved.

ryfu-msft (approved):

PR #461, opened 2023-10-24
Add info command to display key information

### LINKED ISSUE #460: Support `wingetcreate --info` option

Similar to `winget --info`, support `wingetcreate --info` to print out key directories and reference links for winget-create client that helps users find information quickly.

_No response_

mdanish-kh:
	- Resolves #460
	
This PR adds support for root options for the CLI and the following:

- [x] Support for adding Table output. There were a couple of open-source options available but they weren't stylistically similar to how it is displayed in winget CLI.
- [x] User setting to specify whether paths displayed should be anonymous. This change is for both paths displayed on info command and manifest location showed at the end for update command
- [x] Help text for root options

@ryfu-msft Can you kindly create these short URLs if you approve of them ☺️

https://aka.ms/winget-create-license → https://github.com/microsoft/winget-create/blob/main/LICENSE
https://aka.ms/winget-create-privacy → https://github.com/microsoft/winget-create/blob/main/PRIVACY.md
https://aka.ms/winget-create-3rdPartyNotice → https://github.com/microsoft/winget-create/blob/main/NOTICE.txt

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/461)

### DIFF
--- a/src/WingetCreateCLI/Commands/InfoCommand.cs
+++ b/src/WingetCreateCLI/Commands/InfoCommand.cs
@@ -26,7 +26,10 @@ namespace Microsoft.WingetCreateCLI.Commands
         /// <returns>Boolean representing success or fail of the command.</returns>
         public override async Task<bool> Execute()
         {
-            CommandExecutedEvent commandEvent = new CommandExecutedEvent();
+            CommandExecutedEvent commandEvent = new CommandExecutedEvent
+            {
+                Command = nameof(InfoCommand),
+            };
 
             DisplayApplicationHeaderAndCopyright();
             Console.WriteLine();
@@ -55,7 +58,7 @@ namespace Microsoft.WingetCreateCLI.Commands
 
         private static void DisplayInfoTable()
         {
-            string logsdirectory = Common.GetPathForDisplay(Path.Combine(Common.LocalAppStatePath, "DiagOutputDir"), UserSettings.AnonymizePaths);
+            string logsdirectory = Common.GetPathForDisplay(Path.Combine(Common.LocalAppStatePath, Constants.DiagnosticOutputDirectoryFolderName), UserSettings.AnonymizePaths);
             string settingsDirectory = Common.GetPathForDisplay(UserSettings.SettingsJsonPath, UserSettings.AnonymizePaths);
             string installerCacheDirectory = Common.GetPathForDisplay(PackageParser.InstallerDownloadPath, UserSettings.AnonymizePaths);
 
@@ -68,10 +71,10 @@ namespace Microsoft.WingetCreateCLI.Commands
             Console.WriteLine();
 
             new TableOutput(Resources.Links_Heading, string.Empty)
-                            .AddRow(Resources.PrivacyStatement_Heading, "https://aka.ms/winget-create-privacy")
-                            .AddRow(Resources.LicenseAgreement_Heading, "https://aka.ms/winget-create-license")
-                            .AddRow(Resources.ThirdPartyNotices_Heading, "https://aka.ms/winget-create-3rdPartyNotices")
-                            .AddRow(Resources.Homepage_Heading, "https://aka.ms/winget-create")
+                            .AddRow(Resources.PrivacyStatement_Heading, Constants.PrivacyStatementUrl)
+                            .AddRow(Resources.LicenseAgreement_Heading, Constants.LicenseUrl)
+                            .AddRow(Resources.ThirdPartyNotices_Heading, Constants.ThirdPartyNoticesUrl)
+                            .AddRow(Resources.Homepage_Heading, Constants.HomePageUrl)
                             .Print();
         }
     }
--- a/src/WingetCreateCLI/Commands/InfoCommand.cs
+++ b/src/WingetCreateCLI/Commands/InfoCommand.cs
@@ -73,7 +73,7 @@ namespace Microsoft.WingetCreateCLI.Commands
             new TableOutput(Resources.Links_Heading, string.Empty)
                             .AddRow(Resources.PrivacyStatement_Heading, Constants.PrivacyStatementUrl)
                             .AddRow(Resources.LicenseAgreement_Heading, Constants.LicenseUrl)
-                            .AddRow(Resources.ThirdPartyNotices_Heading, Constants.ThirdPartyNoticesUrl)
+                            .AddRow(Resources.ThirdPartyNotices_Heading, Constants.ThirdPartyNoticeUrl)
                             .AddRow(Resources.Homepage_Heading, Constants.HomePageUrl)
                             .Print();
         }
--- a/src/WingetCreateCLI/Commands/InfoCommand.cs
+++ b/src/WingetCreateCLI/Commands/InfoCommand.cs
@@ -1,46 +1,50 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
 
-namespace Microsoft.WingetCreateCLI
+namespace Microsoft.WingetCreateCLI.Commands
 {
     using System;
     using System.IO;
+    using System.Threading.Tasks;
     using CommandLine;
     using Microsoft.WingetCreateCLI.Logging;
     using Microsoft.WingetCreateCLI.Properties;
+    using Microsoft.WingetCreateCLI.Telemetry;
+    using Microsoft.WingetCreateCLI.Telemetry.Events;
     using Microsoft.WingetCreateCore;
+    using Microsoft.WingetCreateCore.Common;
 
     /// <summary>
-    /// Command line options for the root command.
+    /// Info command to display general information regarding the tool.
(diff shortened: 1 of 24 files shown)


src/WingetCreateCLI/Program.cs
```diff
@@ -48,25 +60,46 @@ private static async Task<int> Main(string[] args)
...
-                DisplayParsingErrors(parserResult as NotParsed<object>);
+                // In case of unsuccessful parsing, check if user tried to run a valid command.
+                bool isBaseCommand = baseCommandsParserResult.TypeInfo.Current.BaseType.Equals(typeof(BaseCommand));
+
+                /* Parse root options.
+                 * Adding '--help' is a workaround to force parser to display the options help text when no arguments are passed.
+                 * This makes rootOptionsParserResult to be a NotParsed object which makes HelpText.AutoBuild print the correct help text.
+                 * This is done since the parser does not print the correct help text on a successful parse.
+                  */
+                ParserResult<RootOptions> rootOptionsParserResult = myParser.ParseArguments<RootOptions>(
```
mdanish-kh, line 77: The reason why this results in a NotParsed object even when I've specified a Help Option in RootOptions.cs is because it seems that the parser specifically reserves the `--help` and `--version` keywords. I've only listed Help Option in RootOptions.cs so that I can print a custom help text for it in the CLI output. I realize there is AutoHelp = true available in HelpText.AutoBuild method but that does not let me customize its helptext message.

ryfu-msft (changes requested):
I love everything about having a `--info` menu similar to winget, but i'm a little hesitant about overriding the command parser to achieve parity. The workaround of appending `--help` also feels weird to me as it seems to be changing the intended behavior of the library and I don't know what consequences that may have on other things. 

I know we should try to align with winget-cli as much as possible but if we can achieve the same end result by simply adding a `wingetcreate info` command instead, that may be an easier solution. That would also make it easier for testing so we could verify what is displayed. Thoughts?

mdanish-kh:
I would've loved to keep `--info` but it looks like anything that I'll do to make that happen will involve me doing a very hacky implementation with the parser. And even then all edge cases may not be covered as was the case with `wingetcreate foobar` here

> I know we should try to align with winget-cli as much as possible but if we can achieve the same end result by simply adding a wingetcreate info command instead, that may be an easier solution. That would also make it easier for testing so we could verify what is displayed. Thoughts?

Sounds good. This will also make it easier in future to implement some common options for all commands (e.g. `--log` , `--open-logs` that are mentioned in a couple of issues here)

src/WingetCreateCLI/Commands/InfoCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+                            .AddRow(Resources.InstallerCache_Heading, installerCacheDirectory)
+                            .Print();
+
+            Console.WriteLine();
+
+            new TableOutput(Resources.Links_Heading, string.Empty)
+                            .AddRow(Resources.PrivacyStatement_Heading, "https://aka.ms/winget-create-privacy")
+                            .AddRow(Resources.LicenseAgreement_Heading, "https://aka.ms/winget-create-license")
+                            .AddRow(Resources.ThirdPartyNotices_Heading, "https://aka.ms/winget-create-3rdPartyNotices")
+                            .AddRow(Resources.Homepage_Heading, "https://aka.ms/winget-create")
```
ryfu-msft, line 74: These can be moved to the Constants.cs file in WingetCreateCore so that they can be reused in the future if needed.

src/WingetCreateCLI/Commands/InfoCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        private static void DisplaySystemInformation()
+        {
+            Logger.DebugLocalized(nameof(Resources.OperatingSystem_Info), Environment.OSVersion.VersionString);
+            Logger.DebugLocalized(nameof(Resources.SystemArchitecture_Info), System.Runtime.InteropServices.RuntimeInformation.OSArchitecture);
+        }
+
+        private static void DisplayInfoTable()
+        {
+            string logsdirectory = Common.GetPathForDisplay(Path.Combine(Common.LocalAppStatePath, "DiagOutputDir"), UserSettings.AnonymizePaths);
```
ryfu-msft, line 58: Move `DiagOutputDir` to Constants.cs in WingetCreateCore. There are a couple other places where `DiagOutputDir` is manually typed in, can you also help us replace those with the constant string once you create it.

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -540,7 +540,7 @@
     <value>More help can be found at: https://aka.ms/winget-create</value>
   </data>
   <data name="PrivacyStatement_HelpText" xml:space="preserve">
-    <value>Privacy statement: https://aka.ms/privacy</value>
+    <value>Privacy statement: https://aka.ms/winget-create-privacy</value>
```
ryfu-msft, line 543:
```suggestion
    <value>Privacy statement: {0}</value>
    <comment>{0} represents the privacy statement url</comment>
```

Make sure to use `string.Format` to add the new privacy url for all places that use this resource string.

src/WingetCreateCLI/Common.cs
```diff
@@ -61,6 +64,39 @@ public static void CleanUpFilesOlderThan(string cleanUpDirectory, int cleanUpDay
             }
         }
 
+        /// <summary>
+        /// Gets the path for display. This will anonymize the path if caller provides the appropriate flag.
+        /// </summary>
+        /// <param name="path">Path to be displayed.</param>
+        /// <param name="substituteEnvironmentVariables">Whether or not to substitute environment variables.</param>
+        /// <returns>Anonymized path or original path.</returns>
+        public static string GetPathForDisplay(string path, bool substituteEnvironmentVariables = true)
```
ryfu-msft, line 73: We should add a unit test to verify the behavior of this.

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/InfoCommand.cs
mdanish-kh: Replaced all instances of raw string `DiagOutputDir` in the latest commit

src/WingetCreateCLI/Common.cs
mdanish-kh: Unit test added under `CommonTests.cs` file

src/WingetCreateTests/WingetCreateTests/UnitTests/CommonTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public class CommonTests
+    {
+        /// <summary>
+        /// Tests the ability to retrieve the path for display purposes.
+        /// </summary>
+        [Test]
+        public void VerifyPathSubstitutions()
+        {
+            string path1 = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) + "\\foo\\bar\\baz";
```
ryfu-msft, line 22: Nit: Make `"\\foo\\bar\\baz"` a local string variable inside this test method it's being for all the paths.

src/WingetCreateCore/Common/Constants.cs
```diff
@@ -28,9 +28,39 @@ public static class Constants
...
+        /// </summary>
+        public const string PrivacyStatementUrl = "https://aka.ms/winget-create-privacy";
+
+        /// <summary>
+        /// Link to the license for the winget-create tool.
+        /// </summary>
+        public const string LicenseUrl = "https://aka.ms/winget-create-license";
+
+        /// <summary>
+        /// Link to the notices file containing third party attributions for the winget-create tool.
```
ryfu-msft, line 52:
```suggestion
        /// Link to the notice file containing third party attributions for the winget-create tool.
```

src/WingetCreateCore/Common/Constants.cs
```diff
@@ -28,9 +28,39 @@ public static class Constants
...
+        /// </summary>
+        public const string ThirdPartyNoticesUrl = "https://aka.ms/winget-create-3rdPartyNotices";
```
ryfu-msft, line 54:
```suggestion
        /// Link to the notices file containing third party attributions for the winget-create tool.
        /// </summary>
        public const string ThirdPartyNoticeUrl = "https://aka.ms/winget-create-3rdPartyNotice";
```

ryfu-msft (changes requested):
Just a couple more nits and we can get this checked in 😄 

I'll start working on getting this aka.ms links ready

ryfu-msft:
Having some issues with some internal builds, will investigate and check this PR in when it starts passing. Thanks for the awesome contribution @mdanish-kh!

ryfu-msft (approved):

Issue #458, opened 2023-10-10
Support creating WinGet Configuration files


denelon:
### Description of the new feature / enhancement

I'd like to be able to have tooling to support creating a WinGet Configuration.

We might start with a PowerShell script or something initially to work out the flow, and it could be part of the Microsoft.WinGet.Create module.

wingetcreate is all about manifests today, but it would be nice to have the ability to specify the creation of configuration files.

The tool would be able to iterate over some process like:

Do you want to add a package?
* Get the ID & source

Do you want to configure the package?
* If a DSC Resource is mapped in the source for the package, suggest that resource.
* If a DSC Resource is not mapped, but one is identified a suggestion could include adding that to the manifest.

Do you want to add a Windows Setting?
* Leverage Microsoft.Windows.Developer

Related to:
* https://github.com/microsoft/winget-cli/issues/3523

denelon:
We could follow the "Tools" directory convention used at winget-pkgs to place the PowerShell script while we iterate.

PR #462, opened 2023-10-25
Prototype creating a WinGet Configuration DSC file in PowerShell

KevinLaMS:

-----
Creating DSC files can be a challenge.
This script tries to make it easier by leveraging winget search to locate the packages, and then write the file out as YAML.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/462)


KevinLaMS:
#458 @Trenly your feedback would be appreciated too.

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
+# This script is a prototype for quickly creating DSC files.
+
+if ($null -eq (Get-InstalledModule -Name Microsoft.Winget.Client))
+{
+  Install-Module Microsoft.Winget.Client
+}
+
+if ($null -eq (Get-InstalledModule -Name powershell-yaml))
+{
+  Install-Module powershell-yaml
```
Trenly, line 10: The Install-Module calls should probably be inside Try-Catch that capture any installation errors and re-throw them as a missing dependency error since Install-Module is not always successful

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
+# This script is a prototype for quickly creating DSC files.
+
+if ($null -eq (Get-InstalledModule -Name Microsoft.Winget.Client))
+{
+  Install-Module Microsoft.Winget.Client
```
Trenly, line 5: Does this still require PS Core (v7)? If so, should probably add a #Requires statement to ensure the script can only run in Powershell 7

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
...
+      }
+      else
+      {
+        $packageSelected = $true 
+      }
+    }
+
+    $selectedPackage = $findResult[$selection]
+
+    $unit = @{"resource" = "Microsoft.WinGet.DSC/WinGetPackage"; "directives" = @{"description" = $selectedPackage.Name; "allowPrerelease" = $true; }; "settings" = @{"id" = $selectedPackage.Id; "source"="winget" }}
```
Trenly, line 49: Shouldn’t the source here come from the package the user selected? In theory it would be possible for them to have selected an msstore package, unless you update the initial search to only query the winget source

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
...
+  }
+  else
+  {
+    Write-Host "No package found matching input criteria." -ForegroundColor DarkYellow
+  }
+}
+
+$fileName = Read-Host "Name of the configuration file (without extension)"
+$filePath = Join-Path -Path (Get-Location) -ChildPath "$($fileName).yaml"
+ConvertTo-Yaml @{"properties"= @{"resources"=$finalPackages; "configurationVersion"= $configurationVersion}} -OutFile $filePath -Force
```
Trenly, line 62: I would suggest to explicitly set the encoding on the output file to UTF-8, just to be sure that the output file is always consistent with the encoding winget uses

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
...
+        Write-Host $packageDetails
+        $index++
+    }
+
+    $selection = -1
+    $packageSelected = $false
+    while (-not($packageSelected))
+    {
+      $selection = [int](Read-Host "Input the number of the package you want to select")
+      if (($selection -gt $findResult.count) -or ($selection -lt 0))
```
mdanish-kh, line 37: `$selection -gt $findResult.count` will incorrectly allow one greater than the last value to be a valid selection with the current logic. I would suggest keeping this check as is but do the following:
- initiating $index from 1 (since it's a user-facing number) 
- change ($selection -lt 0) to `($selection -lt 1)`
- change `$selectedPackage = $findResult[$selection]` to be `$selectedPackage = $findResult[$selection - 1]`

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
...
+  else
+  {
+    Write-Host "No package found matching input criteria." -ForegroundColor DarkYellow
+  }
+}
+
+$fileName = Read-Host "Name of the configuration file (without extension)"
+$filePath = Join-Path -Path (Get-Location) -ChildPath "$($fileName).yaml"
+ConvertTo-Yaml @{"properties"= @{"resources"=$finalPackages; "configurationVersion"= $configurationVersion}} -OutFile $filePath -Force
+Write-Host "Configuration file created at: $($filePath)" -ForegroundColor Green
```
mdanish-kh, line 63: Before we print this out, would it be a good idea to run `winget configure validate` on the generated file?

We can add a check if user does not have `winget` on their machine, or even a previous version of winget that does not have `winget configure`

```
if (Get-Command 'winget configure' -ErrorAction SilentlyContinue) { winget configure validate $filePath }
```

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,63 @@
...
+  
+  if ($findResult.count -ne 0)
+  {
+    $index=0
+    foreach ($package in $findResult)
+    {
+        $packageDetails = "[$($index)] $($package.Name) | $($package.Id) | $($package.Version)"
+        Write-Host $packageDetails
+        $index++
+    }
```
Trenly, line 30: This can actually be achieved using the pipeline, and might be cleaner
```powershell
$findResult | ForEach-Object {$i=1} {Add-Member -InputObject $_ -NotePropertyName Index -NotePropertyValue $i; $i++ }
$findResult | Select-Object -Property Index,Name,Id,Version
```

src/prototype/WingetCreateMakeDSC.ps1
Trenly: With my above suggestion adding the index as a property to each package, this would also be simplified
* `if $selection -notin $findResult.Index` can be used to check if the index is in range
* `$selectedPackage = $findResult.Where({$_.Index -eq $selection})` can be used to set the selected package

src/prototype/WingetCreateMakeDSC.ps1
KevinLaMS: Great idea. Completed.

src/prototype/WingetCreateMakeDSC.ps1
KevinLaMS: I went with @mdanish-kh  proposal. I played around with Trenly's but I fell over.  :)

src/prototype/WingetCreateMakeDSC.ps1
KevinLaMS: Done

src/prototype/WingetCreateMakeDSC.ps1
KevinLaMS: I wanted to get an update out, so I skipped this one for now.

src/prototype/WingetCreateMakeDSC.ps1
Trenly: I'll take a look again and might make a PR to your fork if it would be beneficial

KevinLaMS:
@Trenly @mdanish-kh 
1) Thanks for the feedback. I have tried to implement.
2) I resolved the threads, but in retrospect, I probably should have let you do that.  Sorry.
3) Menace me some more, if I messed up your recommendations.

src/prototype/WingetCreateMakeDSC.ps1
KevinLaMS: I tried to address this.

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+    #Winget currently has 2 sources. If the ID contains a period, we will assume winget.
+    #otherwise it is the MSSTORE.  We are not accounting for private REPOs at this time.
+    If ($selectedPackage.Id -like "*.*") {
+      $source="winget"
+    } else {
+      $source="msstore"
+    }
+ 
+    $unit = @{"resource" = "Microsoft.WinGet.DSC/WinGetPackage"; "directives" = @{"description" = $selectedPackage.Name; "allowPrerelease" = $true; }; "settings" = @{"id" = $selectedPackage.Id; "source"=$source }}
+    $tempvar = $finalPackages.Add($unit)
```
mdanish-kh, line 83:
```suggestion
    [void]$finalPackages.Add($unit)
```

nit: Instead of assigning to an unused variable to discard the index printed to stdout, you can cast the result to void.

mdanish-kh (commented):
@KevinLaMS 

> I resolved the threads, but in retrospect, I probably should have let you do that. Sorry.

GitHub allows only the PR author or users with write access to resolve threads, so you did the right thing ☺️
   
> Menace me some more, if I messed up your recommendations.

Did some more menacing😄Most are nits that we can choose to leave and iterate upon them after merging in this PR

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
+# This script is a prototype for quickly creating DSC files.
+
+#Powershell 7 Required
+$hostdata=host
+if ($hostdata.version.major -lt 7) {
```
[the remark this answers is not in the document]
mdanish-kh: Alternatively, we can just add 
```
#Requires -Version 7.0
```

to the top of the script and let pwsh automatically handle the error message. (though we wouldn't get our custom message that way)

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+        $packageSelected = $true 
+      }
+    }
+
+    $selectedPackage = $findResult[$selection - 1] 
+
+    #Specify Source 
+    #Winget currently has 2 sources. If the ID contains a period, we will assume winget.
+    #otherwise it is the MSSTORE.  We are not accounting for private REPOs at this time.
+    If ($selectedPackage.Id -like "*.*") {
```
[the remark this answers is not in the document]
Trenly: I've got an even better solution; Will raise a PR against this branch

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+  }
+  else
+  {
+    Write-Host "No package found matching input criteria." -ForegroundColor DarkYellow
+  }
+}  while ($(Read-Host "Would you like to add another package? [y/n]") -eq 'y')
+
+Write-host
+$fileName = Read-Host "Name of the configuration file (without extension)"
+$filePath = Join-Path -Path (Get-Location) -ChildPath "$($fileName).winget"
```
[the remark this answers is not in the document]
Trenly: I believe there is work going on over at cli to allow double clicking .winget files for automatic install

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+
+ConvertTo-Yaml @{"properties"= @{"resources"=$finalPackages; "configurationVersion"= $configurationVersion}} -OutFile $filePath -Force
```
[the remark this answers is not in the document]
Trenly: Fixing as part of separate pr into this branch

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+  #Pass the exception 
+  throw [System.Net.WebException]::new("Error retrieving powershell module: powershell-yaml. Check that you have installed the Windows Package Manager modules correctly.", $_.Exception)
+  #bugbug is this good enough?
+  }
+}
+
+[System.Collections.ArrayList]$finalPackages = @()
+$configurationVersion = "0.2.0" 
+
+$continue = $true
```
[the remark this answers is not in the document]
Trenly: Fixing as part of separate pr into this branch

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
+# This script is a prototype for quickly creating DSC files.
+
+#Powershell 7 Required
+$hostdata=host
+if ($hostdata.version.major -lt 7) {
+  Write-host "This script requires powershell 7. You can update powershell by typing winget install Microsoft.Powershell." -ForegroundColor red
+  [Environment]::Exit(1)
```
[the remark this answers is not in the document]
Trenly: Fixing as part of separate pr into this branch

src/prototype/WingetCreateMakeDSC.ps1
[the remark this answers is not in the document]
mdanish-kh: Ah, I see. I thought that was only for `.wingetdev` files. Read the [PR](https://github.com/microsoft/winget-cli/pull/3860) description again and it seems `.winget` is also supported

src/prototype/WingetCreateMakeDSC.ps1
```diff
@@ -0,0 +1,111 @@
...
+if ($null -eq (Get-InstalledModule -Name powershell-yaml))
+{
+  try {
+    Install-Module powershell-yaml
+  } catch {
+  #Pass the exception 
+  throw [System.Net.WebException]::new("Error retrieving powershell module: powershell-yaml. Check that you have installed the Windows Package Manager modules correctly.", $_.Exception)
+  #bugbug is this good enough?
+  }
+}
```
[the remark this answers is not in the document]
Trenly: Will address as part of my PR

ryfu-msft (approved):

PR #480, opened 2023-11-21
Add support for creating and updating locales

### LINKED ISSUE #78: Simplify localization for existing packages
# New commands and arguments to help with localization

As someone who prefers a language other than American English (en-US) I want to be able to add or modify localized content to manifests so I can read descriptions in my preferred language.

Two new commands: `wingetcreate new-locale` and `wingetcreate update-locale`

## `wingetcreate new-locale`

`wingetcreate new-locale <PackageIdentifier>`
Targets the latest version, and asks the user for the locale they would like to provide. If the locale is already present, the user should be prompted to use `wingetcreate update-locale`.

If the locale is not present then it provides the default locale for reference so the user can interactively provide the localized content for an additional locale.

## `wingetcreate update-locale`

`wingetcreate update-locale <PackageIdentifier>`
Targets the latest version, and asks the user for the locale they would like to provide. If the locale is not already present, the user should be prompted to use `wingetcreate new-locale`.

If the locale is present then it provides the default locale and the current localized content so the user can interactively provide modified localized content for the locale.

## Additional arguments

`--locale <locale>`: Specify the locale to add or update.
`--reference-locale <locale>`: Specify the locale for comparison other than the default locale.
`--version <version>`: Specify the version for an earlier version of the package .

mdanish-kh:
	- Resolves #78

This PR adds two new commands new-locale and update-locale following the specifications listed in issue #78.

### Locale comparisons

For locale comparions, I've used RegionInfo Class (reference: https://learn.microsoft.com/en-us/dotnet/api/system.globalization.regioninfo?view=net-7.0&redirectedfrom=MSDN). The benefit here is that it validates the locale format and also treats locales such as `en-US`, `en-USA`, `en-us` etc as same which prevents duplicate locales added in the manifest.

~~### Command flow shift~~

~~For now, I've only added supporting for shifting from new-locale to update-locale and vice-versa when `--locale` argument is provided. My thinking here is that since both the commands support creating/updating multiple locales, what if the user enters a non-existing locale after updating multiple locales? User may expect the generated PR to contain both updated and newly created locales, but I've not handled that case here.~~

~~Let me know if I should add the ability to prompt for flow change whenever an invalid locale is input by the user (or only the first time user inputs an invalid locale?).~~  Removed (see https://github.com/microsoft/winget-create/pull/480#discussion_r1418146150)

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/480)


src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,326 @@
...
+                if (referenceLocaleManifest == null)
+                {
+                    Logger.ErrorLocalized(nameof(Resources.ReferenceLocaleNotFound_Error));
+                    return null;
+                }
+            }
+            else
+            {
+                referenceLocaleManifest = originalManifests.DefaultLocaleManifest.ToLocaleManifest();
+            }
```
ryfu-msft, line 264: This check should have already happened before we call this method. If we change to my suggestions where we are passing in the reference locale, the logic for determining the reference locale can be removed.

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,326 @@
...
+
+                var value = fromManifest.GetType().GetProperty(propertyName).GetValue(fromManifest);
+                if (value != null)
+                {
+                    toManifest.GetType().GetProperty(propertyName).SetValue(toManifest, value);
+                }
+            }
+        }
+
+        private List<LocaleManifest> ParseArgumentsAndGenerateLocales(Manifests originalManifests)
```
ryfu-msft, line 239: I think this method does a lot. I think we should do a couple things to simplify it a bit. 

If a reference locale is provided, we should validate that first then retrieve that locale as our reference locale. If none is provided, then our reference locale should just be our defaultLocaleManifest.

Then this method can be simplified so that it only prompts and generates a single locale manifest:

`private LocaleManifest GenerateLocaleManifest(LocaleManifest referenceLocale)`

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,326 @@
...
+            List<string> defaultPromptPropertiesForNewLocale = new()
+            {
+                nameof(LocaleManifest.PackageLocale),
+                nameof(LocaleManifest.PackageName),
+                nameof(LocaleManifest.Publisher),
+                nameof(LocaleManifest.License),
+                nameof(LocaleManifest.ShortDescription),
+            };
+
+            bool localeArgumentUsed = false;
```
ryfu-msft, line 277: Since the new command is always interactive, is there any benefit to having this as an argument? Maybe we should just prompt the user for this to simplify things.

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,326 @@
...
+                else
+                {
+                    LocaleHelper.PromptAndSetLocaleProperties(newLocaleManifest, defaultPromptPropertiesForNewLocale, originalManifests);
+                }
+
+                Console.WriteLine();
+                if (Prompt.Confirm(Resources.AddAdditionalLocaleProperties_Message))
+                {
+                    // Get optional properties that have not been prompted before.
+                    var optionalProperties = LocaleHelper.GetUnPromptedLocalePropertyNames(newLocaleManifest, defaultPromptPropertiesForNewLocale);
```
ryfu-msft, line 310:
```suggestion
                    var optionalProperties = LocaleHelper.GetOptionalLocalePropertyNames(newLocaleManifest, defaultPromptPropertiesForNewLocale);
```

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                                Console.WriteLine();
+                                return await command.Execute();
+                            }
+                        }
+                    }
+                    catch (ArgumentException)
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.InvalidLocale_ErrorMessage));
+                        return false;
+                    }
```
ryfu-msft, line 142: Remove, we don't support switching flows from update -> new

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+            }
+            catch (Octokit.RateLimitExceededException)
+            {
+                Logger.ErrorLocalized(nameof(Resources.RateLimitExceeded_Message));
+                return false;
+            }
+
+            if (!string.IsNullOrEmpty(exactId))
+            {
+                this.Id = exactId;
```
ryfu-msft, line 103: Move inside big try-finally statement so that we still report telemetry.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                    return false;
+                }
+
+                return await Task.FromResult(commandEvent.IsSuccessful = true);
+            }
+            catch (Octokit.NotFoundException e)
+            {
+                Logger.ErrorLocalized(nameof(Resources.Error_Prefix), e.Message);
+                Logger.ErrorLocalized(nameof(Resources.OctokitNotFound_Error));
+                return false;
```
ryfu-msft, line 183: Move this catch statement to where the exception will be thrown.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                InstallerManifest = null,
+            };
+
+            List<string> defaultPromptPropertiesForUpdateLocale = new()
+            {
+                nameof(LocaleManifest.PackageName),
+                nameof(LocaleManifest.Publisher),
+                nameof(LocaleManifest.License),
+                nameof(LocaleManifest.ShortDescription),
+            };
```
ryfu-msft, line 236: How did you come up with these defaults? Since we are dealing with locale, I would assume almost every property would need to be updated. An alternative could be to prompt for every field that contains an existing value except the ones that can't be changed (PackageId, PackageVersion, etc.).

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                    selectedLocale = PromptAllLocalesAsMenuSelection(originalManifests);
+                }
+
+                bool isDefaultLocale = false;
+
+                if (selectedLocale is LocaleManifest localeManifest)
+                {
+                    Logger.DebugLocalized(nameof(Resources.FieldSetToValue_Message), nameof(LocaleManifest.PackageLocale), localeManifest.PackageLocale);
+                    LocaleHelper.PromptAndSetLocaleProperties(localeManifest, defaultPromptPropertiesForUpdateLocale);
+                    updatedLocales.LocaleManifests.Add(localeManifest);
```
ryfu-msft, line 259: You will probably want to add the localeManifest to `updatedLocales` once the user has completed filling out the prompts which includes the additional locale properties. That will make things below a lot simplier.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                    if (isDefaultLocale)
+                    {
+                        PromptOptionalProperties(updatedLocales.DefaultLocaleManifest, LocaleHelper.GetUnPromptedLocalePropertyNames(updatedLocales.DefaultLocaleManifest, defaultPromptPropertiesForUpdateLocale));
+                    }
+                    else
+                    {
+                        LocaleManifest manifest = (LocaleManifest)selectedLocale;
+                        PromptOptionalProperties(manifest, LocaleHelper.GetUnPromptedLocalePropertyNames(manifest, defaultPromptPropertiesForUpdateLocale));
+                    }
+                }
```
ryfu-msft, line 281: Since GetUnpromptedLocalePropertyNames takes in an anonymous parameter, do we need this if-else statement separation between defaultLocale and locale? Ideally the method should be able to handle both manifest types so that there's less repeated code.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                TelemetryManager.Log.WriteEvent(commandEvent);
+            }
+        }
+
+        private static object PromptAllLocalesAsMenuSelection(Manifests originalManifests)
+        {
+            Dictionary<string, object> localeMap = new Dictionary<string, object>
+            {
+                { originalManifests.DefaultLocaleManifest.PackageLocale + " (" + Resources.DefaultLocale_MenuItem + ")", originalManifests.DefaultLocaleManifest },
+            };
```
ryfu-msft, line 196:
```suggestion
            Dictionary<string, object> localeMap = new Dictionary<string, object>
            {
                { originalManifests.DefaultLocaleManifest.PackageLocale + " (" + Resources.DefaultLocale_MenuItem + ")", originalManifests.DefaultLocaleManifest },
            };
```

use string.format or string interpolation here.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                Logger.ErrorLocalized(nameof(Resources.OctokitNotFound_Error));
+                return false;
+            }
+            finally
+            {
+                TelemetryManager.Log.WriteEvent(commandEvent);
+            }
+        }
+
+        private static object PromptAllLocalesAsMenuSelection(Manifests originalManifests)
```
ryfu-msft, line 191: This method should only take in what we need so the parameters should just be DefaultLocaleManifest and LocaleManifests.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+                    {
+                        Logger.ErrorLocalized(nameof(Resources.InvalidLocale_ErrorMessage));
+                        return false;
+                    }
+                }
+
+                Manifests updatedLocales = this.PromptAndUpdateExistingLocales(originalManifests);
+                Console.WriteLine();
+                EnsureManifestVersionConsistency(originalManifests);
+                DisplayUpdatedLocales(updatedLocales);
```
ryfu-msft, line 148: Do the updated Locales ever get assigned back to the originalManifests before they get outputted?

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,291 @@
...
+            foreach (var locale in originalManifests.LocaleManifests)
+            {
+                localeMap.Add(locale.PackageLocale, locale);
+            }
+
+            string selectedLocale = Prompt.Select(Resources.SelectExistingLocale_Message, localeMap.Keys.ToList());
+            return localeMap[selectedLocale];
+        }
+
+        private static void DisplayUpdatedLocales(Manifests manifest)
```
ryfu-msft, line 207: It's not clear why a manifest would be considered 'updated' in this method. The only difference is that you are checking if its null. I'd rather just call the contents of this method in the code above rather than hiding it inside another private method.

ryfu-msft (changes requested):
Apologies for the delay 😅 This is a really cool feature so it took me some time to get through all of it since we don't have very good tests for prompting.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
mdanish-kh: I made these defaults for new-locale for the user to quickly create a very basic/barebone locale manifest, and then prompt if user wants to add additional fields. I made those properties default that are marked as "required" in case of a default locale manifest.

For the update-locale process, it might be more intuitive to prompt for all properties, as it wouldn't be evident beforehand which properties the user is likely to update. Users can also quickly accept values for other properties that don't interest them.

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
mdanish-kh: I'm not sure I follow here. If you're talking about whether we should have `--locale` as an argument or not, then I would say that I tried to remain as close to the original feature request issue as possible where it was requested. As a user, I would like the flexibility to quickly enter a locale in the command and skip a prompt. I don't have strong feelings about it though, if you feel the other way, I can remove `--locale`

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
ryfu-msft: DefaultLocale has `moniker` and Locale doesn't? That feels weird to me, they should be identical...

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
mdanish-kh: updatedLocales would contain all the locales the user has updated through the update-locale command. This method generates a preview of those "updated" locales. Since we don't know whether the updated locale would be a DefaultLocale or LocaleManifest before-hand, we null-check for its existence and print out that locale.

I can refactor this out to remove the method and just call the contents above, but since NewLocaleCommand also has a similar method `DisplayGeneratedLocales` , I would rather keep its counterpart for consistency since the overall flow is quite similar

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
mdanish-kh: That would require converting the DefaultLocale to a LocaleManifest first and then only adding the manifest with `updatedLocales.LocalManifests.Add()`. Given concerns with the type conversion causing us to lose properties (Moniker field), resolving https://github.com/microsoft/winget-create/pull/480#discussion_r1419714365 seems to be a prerequisite for this task.

src/WingetCreateCLI/LocaleHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /// <summary>
+        /// Checks whether package locale already exists in the default locale manifest or one of the locale manifests and returns the matching manifest.
+        /// This function throws an exception if the locale string is in an invalid format.
+        /// </summary>
+        /// <param name="locale">The package locale string to check.</param>
+        /// <param name="originalManifests">The base manifests to check against.</param>
+        /// <returns>An object representing the matching locale manifest.</returns>
+        public static object GetMatchingLocaleManifest(string locale, Manifests originalManifests)
```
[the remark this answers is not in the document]
mdanish-kh: Added a new method DoesLocaleManifestExist and updated existing calls to GetMatchingLocaleManifest to use the new method where we only want to know whether the locale exists or not

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
mdanish-kh: Removed support for changing command flows in the latest commit. Added error message to say to user "Use the update-locale command to update an existing locale manifest"/ "Use the new-locale command to generate a new locale manifest"

mdanish-kh (commented):
I've resolved many of the review comments but left the ones out that require having https://github.com/microsoft/winget-create/pull/480#discussion_r1419714365 resolved first

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
ryfu-msft: Nevermind, I did not realize but yes default locale differs from locale so that we can use the moniker for searching packages. It is okay as is.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,296 @@
...
+                InstallerManifest = null,
+            };
+
+            List<string> defaultPromptPropertiesForUpdateLocale = new()
+            {
+                nameof(LocaleManifest.PackageName),
+                nameof(LocaleManifest.Publisher),
+                nameof(LocaleManifest.License),
+                nameof(LocaleManifest.ShortDescription),
+            };
```
ryfu-msft, line 241: I think having a default set of properties to prompt makes sense for a new locale, but for an update we should just show all of them. My main reason is that there will most definitely be existing locale properties that need to be updated and I'd rather not hide them. The other thing is that it may help reduce some of the complexity of prompting the user for additional properties.

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,296 @@
...
+                DisplayDefaultLocaleManifest(manifest.DefaultLocaleManifest);
+            }
+
+            if (manifest.LocaleManifests != null)
+            {
+                DisplayLocaleManifests(manifest.LocaleManifests);
+            }
+        }
+
+        private Manifests PromptAndUpdateExistingLocales(Manifests originalManifests)
```
ryfu-msft, line 226: Lets just return a tuple of the DefaultLocale and the list of locale manifests. I would also rename the parameter from `originalManifests` to `manifests` and add a method summary comment above it so it clarifies what the input and output is. My main concern is that someone will use this method thinking that the output contains the updated manifests when in fact it is only a partial manifest used solely for displaying  to console.

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+
+                return await Task.FromResult(commandEvent.IsSuccessful = true);
+            }
+            finally
+            {
+                TelemetryManager.Log.WriteEvent(commandEvent);
+            }
+        }
+
+        private static void DisplayGeneratedLocales(List<LocaleManifest> newLocales)
```
ryfu-msft, line 261: make non static

src/WingetCreateCLI/Commands/NewLocaleCommand.cs
```diff
@@ -0,0 +1,328 @@
...
+        {
+            Logger.DebugLocalized(nameof(Resources.GenerateNewLocalePreview_Message));
+            foreach (var localeManifest in newLocales)
+            {
+                Logger.InfoLocalized(nameof(Resources.LocaleManifest_Message), localeManifest.PackageLocale);
+                Console.WriteLine(localeManifest.ToYaml(true));
+            }
+        }
+
+        private static void PopulateLocalePropertiesWithDefaultValues(LocaleManifest reference, LocaleManifest target, List<string> properties)
```
ryfu-msft, line 271: make non static

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,296 @@
...
+            foreach (var locale in localeManifests)
+            {
+                localeMap.Add(locale.PackageLocale, locale);
+            }
+
+            string selectedLocale = Prompt.Select(Resources.SelectExistingLocale_Message, localeMap.Keys.ToList());
+            return localeMap[selectedLocale];
+        }
+
+        private static void DisplayUpdatedLocales(Manifests manifest)
```
ryfu-msft, line 212: make non static

src/WingetCreateCLI/Commands/UpdateLocaleCommand.cs
```diff
@@ -0,0 +1,296 @@
...
+
+                return await Task.FromResult(commandEvent.IsSuccessful = true);
+            }
+            finally
+            {
+                TelemetryManager.Log.WriteEvent(commandEvent);
+            }
+        }
+
+        private static object PromptAllLocalesAsMenuSelection(DefaultLocaleManifest defaultLocale, List<LocaleManifest> localeManifests)
```
ryfu-msft, line 196: Make non static

ryfu-msft (approved):
Super awesome, thanks for working on this!

PR #485, opened 2023-11-26
Improvements to replace argument

### LINKED ISSUE #478: Add replace argument to `submit`

Like added in #277 in my mind it would only be logical to also add the argument to the `submit` command.
I just found a manifest with incorrect information, used the update command with the replace parameter, then corrected and tested the manifest files and now had them submitted with wingetcreate submit. But submit doesn't support a the replace parameter and therefore just created a new PR which made the replace parameter in update have no effect.
If the replace parameter for some reason won't be added to the submit command at least the --submit parameter should be made mandatory if --replace is called or --replace should automatically contain --submit.

**Expected behavior**
winget update [...] -r -> winget submit [...] -> new PR created and old PR requested to be removed/reverted

**Actual behavior**
winget update [...] -r -> winget submit [...] -> new PR created

**OS**
Windows 11 x64 23H2

**wingetcreate version**
WingetCreateCLI 1.5.7.0

mdanish-kh:
	- Resolves #478

### Changes

- Added support for `--replace` for submit command.

- To avoid confusion, I've updated the behavior of update command to fail if submission arguments are not provided with `--submit`. 

Since there isn't good support from the parser library for handling option dependency (ref https://github.com/commandlineparser/commandline/issues/3), I had to handle this manually (i think it's okay since this is only applicable for update command and there could only ever be a handful of arguments that have dependency on --submit).
Now we can decide if we want this to be a "warning" behavior or a "blocking" behavior. I decided to go with a blocking behavior since users are likely to cancel the running update command or run submit command with --replace afterwards if they actually meant to use the replace behavior. Having it fail early makes user quickly recognize mistake and re-run the command with or without --replace.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/485)


src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -361,7 +361,15 @@ private async Task<PullRequest> SubmitPRAsync(string packageId, string version,
...
-                    await this.DeletePackageManifest(repo.Id, packageId, replaceVersion, newBranchName);
+                    try
+                    {
+                        await this.DeletePackageManifest(repo.Id, packageId, replaceVersion, newBranchName);
+                    }
+                    catch (NotFoundException)
+                    {
+                        // If the path to the manifest folder being replaced does not exist, do nothing.
+                        // This can happen with submit command when user submits a new package and path to its version directory does not exist yet.
+                    }
```
ryfu-msft, line 372: We should be explicitly checking whether the folder being replaced exists or not rather than using a try-catch exception to handle that scenario.

src/WingetCreateCLI/Commands/UpdateCommand.cs
```diff
@@ -138,6 +138,22 @@ public override async Task<bool> Execute()
...
+                    Logger.ErrorLocalized(nameof(Resources.SubmitFlagRequired_ErrorMessage));
+                    Console.WriteLine();
+
+                    foreach (string argument in argumentsRequiringSubmit)
+                    {
+                        Logger.Error($"--{argument.ToLower()}");
+                    }
+
+                    return false;
+                }
```
ryfu-msft, line 155: I think this would be better to simply check the condition:
`(!this.SubmitToGitHub && (!string.IsNullOrEmpty(this.PRTitle) || this.Replace))`

Then only warn the user and make it not blocking like so:

`Submit arguments were provided. Did you forget to include the --submit, -s flag?`

Our current behavior is to not make it blocking if the -s flag is not provided and continue outputting an updated manifest so we should adhere to that. Your check and warning would be an improvement though so that we can notify the user if they forget an arg. 

Be sure to also add a unit test to verify that this warning message is outputted. You can look at the existing examples in WingetCreateTests/UnitTests/UpdateCommandTests.cs to see how we check if a string output is correctly shown.

ryfu-msft (changes requested):

src/WingetCreateCore/Common/GitHub.cs
mdanish-kh: Made the flow similar to update command where we validate replace arguments prior to going into GitHub submit flow.

src/WingetCreateCLI/Commands/UpdateCommand.cs
mdanish-kh: Made the behavior non-blocking and added a unit test for checking the output

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -212,6 +212,62 @@ public async Task UpdateFailsWithUnmatchedPackages()
...
+                await command2.Execute();
+            }
+            catch (Exception)
+            {
+                // Expected exception
+            }
+
+            result = this.sw.ToString();
+            Assert.That(result, Does.Contain(Resources.SubmitFlagMissing_Warning), "Submit arguments provided without submit flag warning should be thrown");
+        }
```
ryfu-msft, line 269: Nit: This can be broken up into two separate unit tests named accordingly. That way you don't have to clear the string writer.

src/WingetCreateCLI/Commands/SubmitCommand.cs
```diff
@@ -115,12 +115,24 @@ private async Task<bool> SubmitManifest()
             {
                 Manifests manifests = new Manifests();
                 manifests.SingletonManifest = Serialization.DeserializeFromPath<SingletonManifest>(expandedPath);
+
+                if (this.Replace && !await this.ValidateReplaceArguments(manifests.SingletonManifest.PackageIdentifier, manifests.SingletonManifest.PackageVersion))
+                {
+                    return false;
+                }
```
ryfu-msft, line 122: Not a suggestion but more of a note to self: 
We should probably remove this condition and provide proper error handling since we only support submitting multifile manifests.

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1237,8 +1237,15 @@
...
   </data>
-  <data name="SubmitFlagRequired_ErrorMessage" xml:space="preserve">
-    <value>Submission arguments used without submit flag. The following options require passing in '-s' / '--submit':</value>
-    <comment>'-s / --submit' refers to a command line switch argument</comment>
+  <data name="SubmitFlagMissing_Warning" xml:space="preserve">
+    <value>Submit arguments were provided. Did you forget to include the --submit, -s flag?</value>
+    <comment>'--submit, -s' refers to a command line switch argument</comment>
+  </data>
+  <data name="ReplacePackageIdDoesNotExist_Error" xml:space="preserve">
+    <value>Replace operation cannot be performed. PackageId '{0}' does not exist in the Windows Package Manager repo. </value>
```
ryfu-msft, line 1245:
```suggestion
    <value>Replace operation cannot be performed. Package identifier '{0}' does not exist in the Windows Package Manager repo. </value>

```

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -212,6 +212,62 @@ public async Task UpdateFailsWithUnmatchedPackages()
...
+            {
+                await command1.Execute();
+            }
+            catch (Exception)
+            {
+                // Expected exception
+            }
+
+            string result = this.sw.ToString();
+            Assert.That(result, Does.Contain(Resources.SubmitFlagMissing_Warning), "Submit arguments provided without submit flag warning should be thrown");
```
ryfu-msft, line 253:
```suggestion
            Assert.That(result, Does.Contain(Resources.SubmitFlagMissing_Warning), "Submit flag missing warning should be shown");

ryfu-msft (changes requested):

src/WingetCreateCLI/Commands/SubmitCommand.cs
mdanish-kh: Add a TODO comment

ryfu-msft (approved):

PR #520, opened 2024-03-10
Support updating DisplayVersion in autonomous update

### LINKED ISSUE #516: Ability to update DisplayVersion in autonomous update

Many packages in the winget repository use a different marketing version than the DisplayVersion they write to the Registry. As such, their manifests include both the `PackageVersion` & `DisplayVersion` field. Many publishers follow a consistent pattern with their PackageVersion & DisplayVersion values. As such `DisplayVersion` can be known for such publishers without manually installing the package. 

The current flow for such packages is to create a PR first using wingetcreate, then manually update the DisplayVersion field in the PR. wingetcreate should offer a way to update the DisplayVersion as part of the autonomous update.

A simple `--display-version` arg wouldn't suffice as DisplayVersion can be different per installer node. There should be a way to provide it for each installer URL with overrides as we have right now for architecture & scope. I'm thinking of the following UX, would love additional thoughts or concerns:

- `wingetcreate update x.x --version 1.2.3 --display-version 4.5.6 --urls <InstallerUrls>...` -> Looks for all installer nodes that have a DisplayVersion value and replaces it with `4.5.6`. Do nothing for those that do not have a previous DisplayVersion value.

- `wingetcreate update x.x --version 1.2.3 --urls "<InstallerUrl1>|1.2" "<InstallerUrl2>|x64|2.3" "<InstallerUrl3>|machine|3.4" "<InstallerUrl4>|x86|user|4.5"` -> Replaces the DisplayVersion of installer nodes that match the arch,scope etc with the provided values. If the existing nodes do not have a DisplayVersion value then do nothing.

- `wingetcreate update x.x --version 1.2.3 --display-version 4.5.6 --urls "<InstallerUrl1>|1.2" "<InstallerUrl2> " "<InstallerUrl3>" "<InstallerUrl4>"` -> Matching installer node for InstallerUrl1 gets DisplayVersion `1.2`, rest will get the value provided with `--display-version` arg. For those existing nodes do not have a DisplayVersion value, do nothing.

mdanish-kh:
	- Resolves #516

### Changes
- Adds `--display-version` CLI arg and support for providing display version with the installer URL
- The value in the installer URL takes precedence over value provided with `--display-version`.
- If a single installer has multiple DisplayVersion entries, we aren't sure which one user is targeting to update. In this case, we only update the first DisplayVersion found.

The thing I struggled the most while making this PR is the correct naming convention to be used. I stayed away from calling the value provided in the installer URL as an "override". Winget-Create does not detect the display-version, so calling it an override didn't make much sense to me. It is extra information user is providing in the URL to update. I went with "Installer URL arguments" as the general term for metadata provided with the installer URL. As such, I had to refactor a bunch of existing stuff. I'm still not sure if that's the correct naming, but I'll leave it up for debate here.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/520)


mdanish-kh:
I think your concern is valid. I like the idea of an input object (https://github.com/microsoft/winget-create/issues/521), but I don't think we can get rid of URL overrides completely. That would be a breaking change for many of current CI/CD's that use this tool. What we may want to do is allow scope & arch as they are, but any additional metadata may need to be specified in an object. I see problems with extending overrides as they are, if I want to add support additional string inputs (not convertible to a known enum value like arch, scope, installerType), then winget-create will have a hard time parsing it with the current implementation. I would like to hear @ryfu-msft's thoughts on this one. I'm not that concerned tho about multiple ways for specifying display version as part of this PR. The both ways function differently, and are not a replica of each other. I believe they offer user flexibility in the sense that if user knows that display version is the same for all installers, then specifying `--display-version` may suffice instead of appending the same version with each URL.

If we don't want to allow this as an override by favoring input object method, maybe we can only take the `--display-version` implementation in this PR

yao-msft:
The DisplayVersion field in winget package's manifest is used for installed package version mapping. We advise not to use this field unless there's a strong need to. Basically, when winget sees one DisplayVersion defined in the manifest, it tries to download all manifest versions of the package and read all DisplayVersion values, then winget will create ranges of DisplayVersion to map the Installed Package's ARP DisplayVersion, then the installed package version will be mapped to a PackageVersion of the winget manifest for upgradability check.

This operation is a bit expensive on the client.

Also adding DisplayVersion to manifest has possibility to break the winget-pkg publish pipeline if manifest authors do not know the relation between ARP DisplayVersion and winget manifest PackageVersion because winget-pkgs pipeline has integrity check on the DisplayVersion ranges.

Details can be found [here](https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23980%20-%20Apps%20and%20Features%20entries%20version%20mapping.md)

mdanish-kh:
> Also adding DisplayVersion to manifest has possibility to break the winget-pkg publish pipeline if manifest authors do not know the relation between ARP DisplayVersion and winget manifest PackageVersion because winget-pkgs pipeline has integrity check on the DisplayVersion ranges.

@yao-msft This PR doesn't add support for adding DisplayVersion, it adds support for updating DisplayVersion if the existing manifest already has a DisplayVersion value. In fact, if previous manifest does not have a DisplayVersion value, then the new args will do nothing. This is done to improve the current flow where if an existing manifest contains DisplayVersion field, then winget-create does nothing and leaves the field unchanged in the update flow. This causes the PR to get flagged with error labels over at winget-pkgs since a duplicate DisplayVersion is not allowed within the same PackageIdentifier. User has to manually update this field after creating a PR. This PR improves the UX and allows the user to specify the value while updating the package manifest.

yao-msft:
> @yao-msft This PR doesn't add support for adding DisplayVersion, it adds support for updating DisplayVersion if the existing manifest already has a DisplayVersion value. In fact, if previous manifest does not have a DisplayVersion value, then the new args will do nothing. This is done to improve the current flow where if an existing manifest contains DisplayVersion field, then winget-create does nothing and leaves the field unchanged in the update flow. This causes the PR to get flagged with error labels over at winget-pkgs since a duplicate DisplayVersion is not allowed within the same PackageIdentifier. User has to manually update this field after creating a PR. This PR improves the UX and allows the user to specify the value while updating the package manifest.


Thanks for the clarification. That's really helpful for reducing errors in winget-pkgs!

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -432,10 +445,29 @@ private static void UpdateInstallerMetadata(Installer existingInstaller, Install
             existingInstaller.PackageFamilyName = newInstaller.PackageFamilyName ?? existingInstaller.PackageFamilyName;
             existingInstaller.NestedInstallerFiles = newInstaller.NestedInstallerFiles ?? existingInstaller.NestedInstallerFiles;
             existingInstaller.Platform = newInstaller.Platform ?? existingInstaller.Platform;
+
+            if (existingInstaller.AppsAndFeaturesEntries != null && newInstaller.AppsAndFeaturesEntries != null)
+            {
+                // New installer will always have a single entry in AppsAndFeaturesEntries
```
ryfu-msft, line 451: nit: Can you add to this comment explaining that the AppsAndFeaturesEntries for the new installer will not be null if there is a display-version argument present.

src/WingetCreateCore/Models/InstallerMetadata.cs
```diff
@@ -46,6 +46,11 @@ public class InstallerMetadata
         /// </summary>
         public Scope? OverrideScope { get; set; }
 
+        /// <summary>
+        /// Gets or sets the display version specified as a CLI arg or an installer argument.
```
ryfu-msft, line 50:
```suggestion
        /// Gets or sets the display version specified as a CLI arg or an installer url argument.

```

src/WingetCreateTests/WingetCreateTests/UnitTests/UpdateCommandTests.cs
```diff
@@ -347,52 +347,51 @@ public async Task BlockUpdateSubmissionsWithNoUpdate()
...
-        public async Task UpdateFailsOverrideWithMultipleScopes()
+        public async Task UpdateFailsWithMultipleDisplayVersions()
         {
             string installerUrl = $"https://fakedomain.com/{TestConstants.TestExeInstaller}";
             (UpdateCommand badCommand, var manifests) =
-                GetUpdateCommandAndManifestData("TestPublisher.ScopeOverride", "1.2.3.4", this.tempPath, new[] { $"{installerUrl}|user|machine" });
+                GetUpdateCommandAndManifestData("TestPublisher.ScopeOverride", "1.2.3.4", this.tempPath, new[] { $"{installerUrl}|3.4|1.2" });
             var failedUpdateManifests = await RunUpdateCommand(badCommand, manifests);
-            ClassicAssert.IsNull(failedUpdateManifests, "Command should have failed due to multiple scope overrides specified for a single installer.");
+            ClassicAssert.IsNull(failedUpdateManifests, "Command should have failed due to multiple display versions specified for a single installer.");
```
ryfu-msft, line 392: If you take my suggestion above, this error output will be different so be sure to update this test as well.

ryfu-msft (changes requested):
Sorry this review is so late, just a couple minor comments but overall great new feature 😄

ryfu-msft (approved):

PR #538, opened 2024-07-11
Fix for Loading Invalid Manifests

The-Running-Dev:
- Added new GetManifest, wrapped in try/catch.
- Updated ParseExeInstallerType to use GetManifest

- [ ] Have you signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-pkgs)?

-----

### DIFF
--- a/src/WingetCreateCore/Common/PackageParser.cs
+++ b/src/WingetCreateCore/Common/PackageParser.cs
@@ -765,12 +765,11 @@ namespace Microsoft.WingetCreateCore
         {
             try
             {
-                ManifestResource rc = new ManifestResource();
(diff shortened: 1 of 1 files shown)


ryfu-msft:
@The-Running-Dev, do you have an example installer that I can use to try out this fix?

The-Running-Dev:
Yes I do, try the XYplorer package. That package is the reason this fix exists.

yao-msft:
I feel the correct fix will be to look into why ManifestResource failed on specific manifest and fix the ManifestResource code instead of loading with XmlDocument unchecked and ignore all exceptions.

guerrerosilvana (approved):

Issue #456, opened 2023-09-28
Support creating manifests as JSON


Trenly:
Trenly I think you are on to something. I went through `wingetcreate` again and added the `AppsAndFeaturesEntries`. `wingetcreate` gives me a YAML file, but my repository only accepts JSON files. Unfortunately, I am stupid and do not know how I would go about converting this to the format that would be accepted by the JSON. Any assistance you could provide would be much appreciated. Below is the relevant part of the YAML file that I would need to be in JSON.

```
AppsAndFeaturesEntries:
- DisplayName: SAP GUI for Windows 8.00 32bit  (Patch 3)
  Publisher: SAP SE
  DisplayVersion: 8.00 Compilation 1
- DisplayName: SAP Business Client 8.00
  Publisher: SAP SE
  DisplayVersion: 8.00 PL0
- DisplayName: SAP PDFPRINT for SAP GUI 8.00
  Publisher: SAP SE
- DisplayName: SAP Business Explorer
  Publisher: SAP SE
  DisplayVersion: "8.00"
- DisplayName: 64Bit RFC Controls
  Publisher: SAP SE
  DisplayVersion: "8.00"
```

_Originally posted by @Smith8154 in https://github.com/microsoft/winget-cli/issues/3674#issuecomment-1739577749_

denelon:
This might be the same as
* #3

PR #593, opened 2025-03-28
Bump manifest schema version to 1.10.0

mdanish-kh:

- Bump manifest schema version to 1.10.0. Models are automatically generated by build step through `Newtonsoft.Json`
- Bump winget-create's minor version to 1.10
- Update tests for serializing / deserializing 1.10 yaml and json manifests. Add resource strings as it's needed by the LocalizationTests for all manifest fields. I'm also not including `Authentication` in any of the tests manifests as the WinGet CLI doesn't support Authentication fields. Pending on verified publishers feature to be enabled.
- Skip E2E tests for JSON because of: 

-----

### DIFF
--- a/src/WingetCreateCLI/Properties/Resources.resx
+++ b/src/WingetCreateCLI/Properties/Resources.resx
@@ -1406,4 +1406,16 @@ Warning: Using this argument may result in the token being logged. Consider an a
   <data name="BranchMergeConflict_Message" xml:space="preserve">
     <value>The forked repository could not be synced with the upstream commits due to a merge conflict. Resolve conflicts manually and try again.</value>
   </data>
+  <data name="AuthenticationType_KeywordDescription" xml:space="preserve">
+    <value>The type of authentication to use for Entra Id authentication</value>
+  </data>
+  <data name="Resource_KeywordDescription" xml:space="preserve">
+    <value>The resource to use for Microsoft Entra Id authentication</value>
+  </data>
+  <data name="MicrosoftEntraIdAuthenticationInfo_KeywordDescription" xml:space="preserve">
+    <value>Information required for Entra Id authentication in private sources</value>
+  </data>
+  <data name="Authentication_KeywordDescription" xml:space="preserve">
+    <value>Authentication information for Entra Id secured private sources</value>
+  </data>
 </root>
--- a/src/WingetCreateCore/Common/GitHub.cs
+++ b/src/WingetCreateCore/Common/GitHub.cs
@@ -19,6 +19,7 @@ namespace Microsoft.WingetCreateCore.Common
     using Microsoft.WingetCreateCore.Models.Installer;
     using Octokit;
     using Polly;
+    using AuthenticationType = Octokit.AuthenticationType;
 
     /// <summary>
     /// Provides functionality for interacting a user's GitHub account.
--- a/src/WingetCreateTests/WingetCreateTests/E2ETests/E2ETests.cs
+++ b/src/WingetCreateTests/WingetCreateTests/E2ETests/E2ETests.cs
@@ -55,11 +55,11 @@ namespace Microsoft.WingetCreateE2ETests
         [TestCase(TestConstants.YamlConstants.TestZipPackageIdentifier, TestConstants.YamlConstants.TestZipManifest, TestConstants.TestZipInstaller, TestConstants.YamlManifestFormat)]
 
         // JSON E2E Tests
-        [TestCase(TestConstants.JsonConstants.TestExePackageIdentifier, TestConstants.JsonConstants.TestExeManifest, TestConstants.TestExeInstaller, TestConstants.JsonManifestFormat)]
-        [TestCase(TestConstants.JsonConstants.TestMsiPackageIdentifier, TestConstants.JsonConstants.TestMsiManifest, TestConstants.TestMsiInstaller, TestConstants.JsonManifestFormat)]
-        [TestCase(TestConstants.JsonConstants.TestMultifileMsixPackageIdentifier, TestConstants.JsonConstants.TestMultifileMsixManifestDir, TestConstants.TestMsixInstaller, TestConstants.JsonManifestFormat)]
-        [TestCase(TestConstants.JsonConstants.TestPortablePackageIdentifier, TestConstants.JsonConstants.TestPortableManifest, TestConstants.TestExeInstaller, TestConstants.JsonManifestFormat)]
-        [TestCase(TestConstants.JsonConstants.TestZipPackageIdentifier, TestConstants.JsonConstants.TestZipManifest, TestConstants.TestZipInstaller, TestConstants.JsonManifestFormat)]
+        [TestCase(TestConstants.JsonConstants.TestExePackageIdentifier, TestConstants.JsonConstants.TestExeManifest, TestConstants.TestExeInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestMsiPackageIdentifier, TestConstants.JsonConstants.TestMsiManifest, TestConstants.TestMsiInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestMultifileMsixPackageIdentifier, TestConstants.JsonConstants.TestMultifileMsixManifestDir, TestConstants.TestMsixInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestPortablePackageIdentifier, TestConstants.JsonConstants.TestPortableManifest, TestConstants.TestExeInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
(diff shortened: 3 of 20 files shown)


src/WingetCreateCore/Common/GitHub.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.WingetCreateCore.Common
     using Microsoft.WingetCreateCore.Models.Installer;
     using Octokit;
     using Polly;
+    using AuthenticationType = Octokit.AuthenticationType;
```
AmelBawa-msft, line 22: nit: is this used?

AmelBawa-msft (approved):

src/WingetCreateCore/Common/GitHub.cs
yao-msft: I feel it's not needed.

src/WingetCreateTests/WingetCreateTests/E2ETests/E2ETests.cs
```diff
@@ -54,12 +54,12 @@ public void Setup()
...
+        [TestCase(TestConstants.JsonConstants.TestMsiPackageIdentifier, TestConstants.JsonConstants.TestMsiManifest, TestConstants.TestMsiInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestMultifileMsixPackageIdentifier, TestConstants.JsonConstants.TestMultifileMsixManifestDir, TestConstants.TestMsixInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestPortablePackageIdentifier, TestConstants.JsonConstants.TestPortableManifest, TestConstants.TestExeInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
+        [TestCase(TestConstants.JsonConstants.TestZipPackageIdentifier, TestConstants.JsonConstants.TestZipManifest, TestConstants.TestZipInstaller, TestConstants.JsonManifestFormat, Ignore = "Fails WinGetUtil.WinGetValidateManifest in schema version 1.10.0+")]
```
yao-msft, line 62: Actually, I don't think the Submit to github tests are applicable to json files. So we can just remove these json tests.

yao-msft (commented):
The wingetutil.ValidateManifest was never designed to validate json manifests in mind. Especially later the winget REST source manifests (the json manifests) differ in format with the yaml manifests. So the parsing logic for json manifests are also different from yaml manifests. Plus, I don't think for a json, we will ever need multifile manifests.

For the time being, we could always serialize as yaml for wingetutil.validatemanifest(), if user specified --format json, then serialize as json as output.

We'll probably deprecate this type of json serializing in the future (when we work on winget rest source support from wingetcreate), at that time it will need to serialize as winget rest source(winget-cli-restsource) compliant json manifest, then we can submit the manifests to a test rest source.

src/WingetCreateCore/Common/GitHub.cs
mdanish-kh: Manifest schema `1.10.0` has a new `AuthenticationType` field which conflicts with the definition from `Octokit` library. This is done to disambiguate the reference for the compiler. I could just use `Octokit.AuthenticationType` in multiple places where this reference throughout this file, but I thought this was cleaner to define it up top since we won't be using the installer manifest schema definition in this file

image

src/WingetCreateTests/WingetCreateTests/E2ETests/E2ETests.cs
mdanish-kh: They are for private or personal GitHub repositories. The feature ask in https://github.com/microsoft/winget-create/issues/456 was to support submitting it to a fork of winget-pkgs. The tests would cover this case to submit it to a GitHub repo that's not `microsoft/winget-pkgs`. The implementation was such that JSON would be blocked to be submitted to `microsoft/winget-pkgs`, but would be allowed to submit it to some other repository (which can be specified under `wingetcreate settings`).


See: 
- https://github.com/microsoft/winget-pkgs-submission-test/pull/5996

and 

https://github.com/microsoft/winget-create/blob/1e78f07d72249c28cff23dfcb2f5ba0db565f3d5/src/WingetCreateCLI/Commands/BaseCommand.cs#L720-L727

mdanish-kh: Though... if we are not supporting submitting JSON _anywhere_, we might as well get rid of those tests. Not sure then what we would decide to do with the `--format json` switch that the CLI supports for `new`, `update`, `show`, `update-locale`, `new-locale` commands. Those will still output json, just that the user won't be able to use wingetcreate to directly submit those anywhere as it would fail the `ValidateManifests()` check

> For the time being, we could always serialize as yaml for wingetutil.validatemanifest(), if user specified --format json, then serialize as json as output.

I like this thought, but I think this will require a larger change. WingetUtil.ValidateManifest() accepts a manifest path, which means what `wingetcreate` would have to do is to always output YAML files, validate it & then add logic post-validation based on the `--format` switch or `ManifestFormat` in settings to convert the output yaml files to JSON. Right now what it does is generate `.json` or `.yaml` files in a directory based on `--format` and pass the path to wingetutil.validatemanifest() and then use the same path for GitHub submission.

All of this to say, we need to decide if doing all this is worth for the use case, especially if we are not seeing any usage of this feature through telemetry...

mdanish-kh:
The test failure reason seems to be the same one that was handled in

- https://github.com/microsoft/winget-create/pull/558

but appears to be broken again


All tests pass for me locally

src/WingetCreateTests/WingetCreateTests/E2ETests/E2ETests.cs
mdanish-kh: Options:

1. Do not support submitting JSON. Remove E2E tests & any other submission tests for json. Would also need to reason if we should then completely remove the `--format` switch for all commands & associated user settings or just fail with a message that we do not support submitting JSON manifests
2. Skip validation for JSON manifests. Have a warning output that says we do not support validating JSON manifests but users would still be able to make the submission
3. Do the larger change of outputting YAML only, validating it, then converting to JSON making sure the YAMLs are cleaned up properly and then allow submission

Let me know what makes the most sense for this PR & the release that we want to cut for supporting 1.10.0 schema.

yao-msft: > They are for private or personal GitHub repositories. The feature ask in #456 was to support submitting it to a fork of winget-pkgs. The tests would cover this case to submit it to a GitHub repo that's not `microsoft/winget-pkgs`. The implementation was such that JSON would be blocked to be submitted to `microsoft/winget-pkgs`, but would be allowed to submit it to some other repository (which can be specified under `wingetcreate settings`).
> 
> See:
> 
> * [WingetCreateE2E.Json.ZipTest 1.2.3.4 (Json) winget-pkgs-submission-test#5996](https://github.com/microsoft/winget-pkgs-submission-test/pull/5996)
> 
> and
> 
> https://github.com/microsoft/winget-create/blob/1e78f07d72249c28cff23dfcb2f5ba0db565f3d5/src/WingetCreateCLI/Commands/BaseCommand.cs#L720-L727

So I just went over the original thread and yes, the user was submitting to winget REST source, where the manifest format is different from yaml manifest for winget-pkgs.

We will be doing "submitting manifest to REST source" feature later, which is different from what current code is producing. So I think we should at least deprecate current code when doing that feature (anyway, this feature is not working with latest wingetutil.dll)

yao-msft: That being said, we could safely remove the tests to push json manifests to github repo. And we can fix the json manifests format (for REST source) later when we work on publishing to REST source in the future.

mdanish-kh: > That being said, we could safely remove the tests to push json manifests to github repo. 

Sounds good, I removed the E2E tests for pushing JSON to GitHub repo

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1406,4 +1406,19 @@ Warning: Using this argument may result in the token being logged. Consider an a
   <data name="BranchMergeConflict_Message" xml:space="preserve">
     <value>The forked repository could not be synced with the upstream commits due to a merge conflict. Resolve conflicts manually and try again.</value>
   </data>
+  <data name="AuthenticationType_KeywordDescription" xml:space="preserve">
+    <value>The type of authentication to use for Entra Id authentication</value>
+  </data>
+  <data name="Resource_KeywordDescription" xml:space="preserve">
```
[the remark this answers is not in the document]
mdanish-kh: I think I remember why I chose to omit this description, but forgot to call this out in the PR or implementation. The logic for outputting these resource strings in the interactive menu doesn't handle cases where there is a collision between property names, which is what we're seeing right now with `Installers > Scope` and `Installers > Authentication > MicrosoftEntraIdAuthentication > Scope` where both have the same property name "Scope". I think I reasoned that users _probably_ won't be using wingetcreate with its interactive menu to add this value but would be manually adding it in the manifest, I just let the description of it be the Installers > Scope one in the initial implementation.

I've updated the code now to handle this specific case separately for now. I don't like this code either :D but added a TODO in the code comments of what would be a bigger change to handle this. Let me know if that sounds reasonable or would you like me to handle this in another way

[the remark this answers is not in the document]
mdanish-kh: Manually verified that the string output is the correct one in a `wingetcreate new` command (or `wingetcreate update --interactive` command)

(below is a GIF)

resourceString

PR #607, opened 2025-06-12
Initial support for DSC v3 in WinGet-Create

AmelBawa-msft:
# ⚙️ Settings DSC resource
## 📄 Get
```shell
PS C:\> wingetcreate dsc settings --get
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"A","name":"B"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}}}
```

## 🖨️ Export
ℹ️ Settings resource Get and Export operation output states are identical.
```shell
PS C:\> wingetcreate dsc settings --export
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"A","name":"B"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}}}
```

## 📝 Set
- Action `Full`: When action is set to Full, the specified settings will be update accordingly and the remaining settings will be set to their default values.
- Action `Partial`: When action is set to Partial, only the specified settings will be updated, and the remaining settings will remain unchanged.
### 🌕 Full
```shell
PS C:\> wingetcreate dsc settings --set '{"settings": { "Telemetry": { "disable": false }}, "action": "Full"}'
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":false},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Full"}
["settings"]
```

### 🌗 Partial
```shell
PS C:\> wingetcreate dsc settings --set '{"settings": { "Telemetry": { "disable": true }}, "action": "Partial"}'
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Partial"}
["settings"]
```

## 🧪 Test
- Action `Full`: When action is set to Full, the specified settings will be tested accordingly, and the remaining settings will be tested against their default values.
- Action `Partial`: When action is set to Partial, only the specified settings will be tested, and the remaining settings will be omitted from the test.
### 🌕 Full
```shell
PS C:\> wingetcreate dsc settings --test '{"settings": { "Telemetry": { "disable": false }}, "action": "Full"}'
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Full","_inDesiredState":false}
["settings"]
```

### 🌗 Partial
```shell
PS C:\> wingetcreate dsc settings --test '{"settings": { "Telemetry": { "disable": false }}, "action": "Partial"}'
{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Partial","_inDesiredState":false}
["settings"]
```

### DIFF
--- a/doc/dsc/settings.md
+++ b/doc/dsc/settings.md
@@ -1,0 +1,49 @@
+# Settings resource
+Manage the settings for Winget-Create
+
+## 📄 Get
+```shell
+PS C:\> wingetcreate dsc settings --get
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}}}
+```
+
+## 🖨️ Export
+ℹ️ Settings resource Get and Export operation output states are identical.
+```shell
+PS C:\> wingetcreate dsc settings --export
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}}}
+```
+
+## 📝 Set
+- Action `Full`: When action is set to Full, the specified settings will be update accordingly and the remaining settings will be set to their default values.
+- Action `Partial`: When action is set to Partial, only the specified settings will be updated, and the remaining settings will remain unchanged.
+### 🌕 Full
+```shell
+PS C:\> wingetcreate dsc settings --set '{"settings": { "Telemetry": { "disable": false }}, "action": "Full"}'
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":false},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Full"}
+["settings"]
+```
+
+### 🌗 Partial
+```shell
+PS C:\> wingetcreate dsc settings --set '{"settings": { "Telemetry": { "disable": true }}, "action": "Partial"}'
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Partial"}
+["settings"]
+```
+
+## 🧪 Test
+- Action `Full`: When action is set to Full, the specified settings will be tested accordingly, and the remaining settings will be tested against their default values.
+- Action `Partial`: When action is set to Partial, only the specified settings will be tested, and the remaining settings will be omitted from the test.
+### 🌕 Full
+```shell
+PS C:\> wingetcreate dsc settings --test '{"settings": { "Telemetry": { "disable": false }}, "action": "Full"}'
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Full","_inDesiredState":false}
+["settings"]
+```
+
+### 🌗 Partial
+```shell
+PS C:\> wingetcreate dsc settings --test '{"settings": { "Telemetry": { "disable": false }}, "action": "Partial"}'
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}},"action":"Partial","_inDesiredState":false}
+["settings"]
+```
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -62,6 +62,11 @@ namespace Microsoft.WingetCreateCLI.Commands
         public static string Extension => Serialization.ManifestSerializer.AssociatedFileExtension;
 
         /// <summary>
+        /// Gets a value indicating whether or not the command requires a GitHub token to be set.
+        /// </summary>
+        public virtual bool RequiresGitHubToken { get; } = true;
+
+        /// <summary>
         /// Gets or sets the GitHub token used to submit a pull request on behalf of the user.
         /// </summary>
         public virtual string GitHubToken { get; set; }
--- a/src/WingetCreateCLI/Commands/BaseCommand.cs
+++ b/src/WingetCreateCLI/Commands/BaseCommand.cs
@@ -62,9 +62,9 @@ namespace Microsoft.WingetCreateCLI.Commands
         public static string Extension => Serialization.ManifestSerializer.AssociatedFileExtension;
 
         /// <summary>
-        /// Gets a value indicating whether or not the command requires a GitHub token to be set.
+        /// Gets a value indicating whether or not the command accepts a GitHub token.
         /// </summary>
-        public virtual bool RequiresGitHubToken { get; } = true;
+        public virtual bool AcceptsGitHubToken { get; } = true;
 
         /// <summary>
         /// Gets or sets the GitHub token used to submit a pull request on behalf of the user.
--- a/src/WingetCreateCLI/Commands/DscCommand.cs
+++ b/src/WingetCreateCLI/Commands/DscCommand.cs
@@ -9,7 +9,6 @@ using System.Linq;
 using System.Threading.Tasks;
 using CommandLine;
 using Microsoft.WingetCreateCLI.Commands.DscCommands;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 /// <summary>
--- a/src/WingetCreateCLI/Commands/DscCommand.cs
+++ b/src/WingetCreateCLI/Commands/DscCommand.cs
@@ -51,6 +51,12 @@ public class DscCommand : BaseCommand
     public string Export { get; set; }
 
     /// <summary>
+    /// Gets or sets a value indicating whether to execute the schema command.
+    /// </summary>
+    [Option("schema", SetName = "SchemaMethod", HelpText = "Command for the Schema flow.")]
+    public bool Schema { get; set; }
+
+    /// <summary>
     /// Executes the dsc command flow.
     /// </summary>
     /// <returns>Boolean representing success or fail of the command.</returns>
@@ -58,7 +64,7 @@ public class DscCommand : BaseCommand
     {
         BaseDscCommand dscCommand;
         var dscScope = this.UnboundArgs.FirstOrDefault()?.ToLowerInvariant() ?? string.Empty;
-        if (dscScope == DscSettingsCommand.CommandName)
+        if (dscScope == "settings")
         {
             dscCommand = new DscSettingsCommand();
         }
@@ -84,6 +90,10 @@ public class DscCommand : BaseCommand
         else if (this.TryParse(this.Export, out input))
         {
             dscCommand.Export(input);
+        }
+        else if (this.Schema)
+        {
+            dscCommand.Schema();
         }
         else
         {
--- a/src/WingetCreateCLI/Commands/DscCommand.cs
+++ b/src/WingetCreateCLI/Commands/DscCommand.cs
@@ -5,6 +5,7 @@ namespace Microsoft.WingetCreateCLI.Commands;
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using CommandLine;
@@ -72,11 +73,9 @@ public class DscCommand : BaseCommand
     {
         await Task.CompletedTask;
 
-        List<BaseDscCommand> dscCommands = [new DscSettingsCommand()];
-        var dscCommand = dscCommands.FirstOrDefault(c => c.CommandName.Equals(this.ResourceName, StringComparison.OrdinalIgnoreCase));
-        if (dscCommand == null)
+        if (!BaseDscCommand.TryCreateInstance(this.ResourceName, out var dscCommand))
         {
-            var availableResources = string.Join(", ", dscCommands.Select(c => c.CommandName));
+            var availableResources = string.Join(", ", BaseDscCommand.GetAvailableCommands());
             Logger.ErrorLocalized(nameof(Resources.DscResourceNameNotFound_Message), this.ResourceName, availableResources);
             return false;
         }
--- a/src/WingetCreateCLI/Commands/DscCommand.cs
+++ b/src/WingetCreateCLI/Commands/DscCommand.cs
@@ -77,7 +77,7 @@ public class DscCommand : BaseCommand
         if (dscCommand == null)
         {
             var availableResources = string.Join(", ", dscCommands.Select(c => c.CommandName));
-            Logger.ErrorLocalized(nameof(Resources.DscResourceNotFound_Message), this.ResourceName, availableResources);
+            Logger.ErrorLocalized(nameof(Resources.DscResourceNameNotFound_Message), this.ResourceName, availableResources);
             return false;
         }
 
@@ -107,7 +107,7 @@ public class DscCommand : BaseCommand
                 }
             }
 
-            Logger.ErrorLocalized(nameof(Resources.DscResourceOperationInvalid_Message));
+            Logger.ErrorLocalized(nameof(Resources.DscResourceOperationNotSpecified_Message));
             return false;
         }
         catch (Exception ex)
--- a/src/WingetCreateCLI/Commands/DscCommand.cs
+++ b/src/WingetCreateCLI/Commands/DscCommand.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using CommandLine;
 using Microsoft.WingetCreateCLI.Commands.DscCommands;
+using Microsoft.WingetCreateCLI.Logging;
 using Newtonsoft.Json.Linq;
 
 /// <summary>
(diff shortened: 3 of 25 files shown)


src/WingetCreateCLI/Commands/DscCommand.cs
```diff
@@ -0,0 +1,117 @@
...
+public class DscCommand : BaseCommand
+{
+    /// <inheritdoc/>
+    public override bool RequiresGitHubToken => false;
+
+    /// <summary>
+    /// Gets or sets the unbound arguments that exist after the positional parameters.
+    /// </summary>
+    [Value(2, Hidden = true)]
+    public IList<string> UnboundArgs { get; set; } = new List<string>();
```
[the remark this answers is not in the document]
mdanish-kh: The input JSON object object can also be a positional argument instead of being an unbound arg, in this case it'll come at position 1 (position 0 will be for commandName). A similar example for this would be the `ReplaceVersion` positional arg in UpdateCommand.cs

https://github.com/microsoft/winget-create/blob/f26390d404319020cb9d67d4d7c27c032e1cc158/src/WingetCreateCLI/Commands/UpdateCommand.cs#L58-L62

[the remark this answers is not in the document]
AmelBawa-msft: Great suggestion, thank you 🔥 I added two properties for the resource name and the input JSON, this also makes the DSC resource JSON look nicer 😉

Technically, we don't expect users to call this API directly, instead the DSC resource file would act as the adapter which the `dsc.exe` command would understand on how to call this tool.

For example, the user types:
```
dsc resource get -r "Microsoft.WinGetCreate/Settings"
```

`dsc.exe` translate it (using the DSC resource json file) to:

```
wingetcreate dsc settings --get
```

src/WingetCreateCLI/Commands/BaseCommand.cs
```diff
@@ -61,6 +61,11 @@ public abstract class BaseCommand
         /// </summary>
         public static string Extension => Serialization.ManifestSerializer.AssociatedFileExtension;
 
+        /// <summary>
+        /// Gets a value indicating whether or not the command requires a GitHub token to be set.
+        /// </summary>
+        public virtual bool RequiresGitHubToken { get; } = true;
```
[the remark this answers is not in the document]
AmelBawa-msft: Gotcha! I like your suggestion `AcceptsGitHubToken`. Updated 🥳

src/WingetCreateCLI/DscResources/microsoft.winget-create.settings.dsc.resource.json
```diff
@@ -0,0 +1,66 @@
...
+    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/bundled/resource/manifest.json",
+    "description": "Allows management of settings state via the DSC v3 command line interface protocol. See the help link for details.",
+    "export": {
+        "executable": "wingetcreate",
+        "input": "stdin",
+        "args": [
+            "dsc",
+            "settings",
+            "--export",
+            "'{}'"
```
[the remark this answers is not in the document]
AmelBawa-msft: Because the input JSON was promoted to a property as you suggested. Now we can simplify this hacky syntax 😉

mdanish-kh (commented):
This looks great!

_(P.S. maybe you can modify the PR description for the new changes to command syntax to help others who will review this PR :D)_

doc/dsc/settings.md
```diff
@@ -0,0 +1,49 @@
...
+PS C:\> wingetcreate dsc settings --export
+{"settings":{"$schema":"https://aka.ms/wingetcreate-settings.schema.0.1.json","Telemetry":{"disable":true},"CleanUp":{"intervalInDays":7,"disable":false},"WindowsPackageManagerRepository":{"owner":"microsoft","name":"winget-pkgs"},"Manifest":{"format":"yaml"},"Visual":{"anonymizePaths":true}}}
+```
+
+## 📝 Set
+- Action `Full`: When action is set to Full, the specified settings will be update accordingly and the remaining settings will be set to their default values.
+- Action `Partial`: When action is set to Partial, only the specified settings will be updated, and the remaining settings will remain unchanged.
+### 🌕 Full
+```shell
+PS C:\> wingetcreate dsc settings --set '{"settings": { "Telemetry": { "disable": false }}, "action": "Full"}'
```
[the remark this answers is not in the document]
mdanish-kh: actuallyy.... this may not be a good suggestion, since I see this wouldn't then work when specifying string values, i.e., the value of action ("Partial", "Full") or ManifestFormat ("yaml", "json")... it could be that the VS settings isn't doing a good job of passing these CLI args and escaping the single quote.. will have to see.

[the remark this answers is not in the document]
mdanish-kh: Deployed the solution & tested in PowerShell. I guess this was VS messing with me

image

src/WingetCreateCLI/Commands/DscCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// <summary>
+    /// Executes the dsc command flow.
+    /// </summary>
+    /// <returns>Boolean representing success or fail of the command.</returns>
+    public override async Task<bool> Execute()
+    {
+        await Task.CompletedTask;
+
+        List<BaseDscCommand> dscCommands = [new DscSettingsCommand()];
```
yao-msft, line 75: Creating a list of commands and use only 1 of them is inefficient. Have a private helper method to create the resource command needed for resource name.

src/WingetCreateCLI/Commands/DscCommands/BaseDscCommand.cs
```diff
@@ -0,0 +1,82 @@
...
+        var resourceObject = new T();
+        return new JObject
+        {
+            ["$schema"] = "http://json-schema.org/draft-07/schema#",
+            ["title"] = this.CommandName,
+            ["type"] = "object",
+            ["properties"] = resourceObject.GetProperties(),
+            ["required"] = resourceObject.GetRequiredProperties(),
+            ["additionalProperties"] = false,
+        };
```
yao-msft, line 71: This does not seem like a correct dsc schema format

src/WingetCreateCLI/DscResources/microsoft.winget-create.settings.dsc.resource.json
```diff
@@ -0,0 +1,62 @@
+{
```
yao-msft, line 1: Like winget-cli dsc, we should generate the resource manifest by running --schema command so we don't need to maintain a separate file for future updates.

src/WingetCreateCLI/DscResources/microsoft.winget-create.settings.dsc.resource.json
```diff
@@ -0,0 +1,62 @@
...
+                "jsonInputArg": "--test",
+                "mandatory": true
+            }
+        ]
+    },
+    "tags": [
+        "WinGetCreate"
+    ],
+    "type": "Microsoft.WinGetCreate/Settings",
+    "version": "1.9.0"
```
yao-msft, line 61: I just realized winget-create has not been updated to latest since 1.9...

src/WingetCreateCLI/Commands/DscCommands/DscSettingsCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+        data.Get();
+        data.Output.InDesiredState = data.Test();
+
+        this.WriteJsonOutputLine(data.Output.ToJson());
+        this.WriteJsonOutputLine(data.DiffJson());
+        return true;
+    }
+
+    /// <inheritdoc/>
+    public override bool Export(JToken input)
```
yao-msft, line 72: seems like input was never used, same for get

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1406,4 +1406,51 @@ Warning: Using this argument may result in the token being logged. Consider an a
...
+    <value>Set the resource state</value>
+  </data>
+  <data name="DscTest_HelpText" xml:space="preserve">
+    <value>Test the resource state</value>
+  </data>
+  <data name="DscExport_HelpText" xml:space="preserve">
+    <value>Get all state instances</value>
+  </data>
+  <data name="DscSchema_HelpText" xml:space="preserve">
+    <value>Execute the Schema command</value>
```
yao-msft, line 1425: Outputs schema of the resource

src/WingetCreateCLI/Models/DscModels/BaseResourceObject.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <summary>
+    /// Converts the current object to a JSON representation.
+    /// </summary>
+    /// <returns>A Json object representing the current object.</returns>
+    public JObject ToJson()
+    {
+        return JObject.FromObject(this, new JsonSerializer
+        {
+            NullValueHandling = NullValueHandling.Ignore,
```
yao-msft, line 51: nit: Null Ignore is also defined in InDesiredState property decoration, do we want individual property to define its own behavior or define the behavior in top level. It seems one of them is redundant.

src/WingetCreateCLI/Models/DscModels/SettingsFunctionData.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        return this.resolvedInputUserSettings;
+    }
+
+    /// <summary>
+    /// Retrieves a deep-cloned JSON representation of the current user settings.
+    /// </summary>
+    /// <returns>A Json object representing the user settings.</returns>
+    public JObject GetUserSettings()
```
yao-msft, line 109: nit: private

src/WingetCreateCLI/Models/DscModels/SettingsResourceObject.cs
```diff
@@ -0,0 +1,61 @@
...
+    public string Action { get; set; }
+
+    /// <inheritdoc/>
+    public override JObject GetProperties()
+    {
+        var baseProperties = base.GetProperties();
+        baseProperties["settings"] = new JObject
+        {
+            ["description"] = Resources.DscResourcePropertyDescriptionSettings,
+            ["type"] = "object",
```
yao-msft, line 44: I thought we are parsing jobject from a string

src/WingetCreateCLI/Models/DscModels/SettingsResourceObject.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// Gets or sets the settings for the resource object.
+    /// </summary>
+    [JsonProperty("settings", Required = Required.Always)]
+    public JObject Settings { get; set; }
+
+    /// <summary>
+    /// Gets or sets the action to be performed on the settings resource object.
+    /// </summary>
+    [JsonProperty("action", NullValueHandling = NullValueHandling.Ignore)]
+    public string Action { get; set; }
```
yao-msft, line 35: "= ActionPartial" to be explicit?

src/WingetCreateCLI/Models/DscModels/SettingsFunctionData.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        this.userSettings ??= UserSettings.ToJson();
+        return (JObject)this.userSettings.DeepClone();
+    }
+
+    /// <summary>
+    /// Writes the current output settings to persistent storage.
+    /// </summary>
+    public void WriteOutput()
```
yao-msft, line 118: nit: this could actually be named "Set" to be more clear

src/WingetCreateCLI/WingetCreateCLI.csproj
```diff
@@ -96,12 +96,20 @@
...
+    <SettingSchema Include="$(ProjectDir)\Schemas\settings.schema.0.1.json" ModelName="Settings" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Models\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="$(ProjectDir)\DscResources\microsoft.winget-create.*.dsc.resource.json">
+      <Link>%(Filename)%(Extension)</Link>
```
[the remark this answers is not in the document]
AmelBawa-msft: The Link here ensures the dsc.resource.json is placed at the root of the target directory, right next to the exe.
Verified that the generated msix package has the resource.dsc.json in the root directory.

src/WingetCreateCLI/Commands/DscCommand.cs
```diff
@@ -0,0 +1,119 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.WingetCreateCLI.Commands;
```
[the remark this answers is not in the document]
AmelBawa-msft: I did notice that we use the more traditional block scoped namespace previously. However, I do prefer using the file scoped namespace style as it is more convenient to read. But I do agree that we should also make it consistent. I will consider this change separately 🙂

src/WingetCreateCLI/Commands/DscCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+    [Option("schema", SetName = "SchemaMethod", HelpText = "DscSchema_HelpText", ResourceType = typeof(Resources))]
+    public bool Schema { get; set; }
+
+    /// <summary>
+    /// Executes the dsc command flow.
+    /// </summary>
+    /// <returns>Boolean representing success or fail of the command.</returns>
+    public override async Task<bool> Execute()
+    {
+        await Task.CompletedTask;
```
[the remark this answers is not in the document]
AmelBawa-msft: The method expects an await since it is async. I can remove the async but then I have to return a Task. The `await Task.CompletedTask` is a common way to workaround the signature requirement (which is inherited from the parent class)

src/WingetCreateCLI/Commands/DscCommand.cs
AmelBawa-msft: I was thinking about a better way to do this, I wanted something like a static inheritance property that does not require initializing an instance to access the value, but that does not exist AFAIK. I am not worried about memory/process performance here as this is a lightweight operation that intentionally prioritize convenience and scalability over performance.

src/WingetCreateCLI/Commands/DscCommands/BaseDscCommand.cs
AmelBawa-msft: Could you please elaborate on what is potentially the issue with this code? I copied it from winget-cli in the below link:

https://github.com/microsoft/winget-cli/blob/26e5eca45967b4949e96883aa9f5f953ceb84518/src/AppInstallerCLICore/Commands/DscComposableObject.cpp#L49-L59

src/WingetCreateCLI/DscResources/microsoft.winget-create.settings.dsc.resource.json
AmelBawa-msft: This file isn’t the schema for the DSC resource; it's the manifest that instructs DSC on how to interface with wingetcreate. While winget-cli does generate this file using the --manifest option, I don’t anticipate changes to it beyond version updates. That’s something we could consider improving in the future.
Regarding the schema itself, I’ve already implemented a --schema option that outputs the schema definition. This is also what the manifest directs DSC to do when the schema command is invoked via dsc resource schema -r ....

src/WingetCreateCLI/DscResources/microsoft.winget-create.settings.dsc.resource.json
AmelBawa-msft: Update version to 1.10.0 after the PR above was merged :)

src/WingetCreateCLI/Commands/DscCommands/DscSettingsCommand.cs
AmelBawa-msft: Get, Set, Test, and Export are methods that will be overridden by DSC commands. While the settings command does not require input for Get and Export, other DSC commands may depend on these inputs.

src/WingetCreateCLI/Models/DscModels/BaseResourceObject.cs
AmelBawa-msft: The extra property here is not needed. Good catch!

src/WingetCreateCLI/Models/DscModels/SettingsResourceObject.cs
AmelBawa-msft: No, the settings property expects a JSON object as input.

src/WingetCreateCLI/Commands/DscCommand.cs
yao-msft: I'm not asking a sophisticated way of creating the instance. Just a helper method with (in any case, you'll have to maintain this mapping somehow)

if (this.ResourceName == SettingsCommand.CommandName)
{
  return new SettingsCommand()
}

//error case

src/WingetCreateCLI/Commands/DscCommands/BaseDscCommand.cs
yao-msft: If the output of settings dsc command schema is complete and correct, you can ignore this comment.

src/WingetCreateCLI/Models/DscModels/SettingsResourceObject.cs
yao-msft: nit: if we don't want default value, type should be "string?"

I know we have not turned on stylecop for this project, but we will one day..

src/WingetCreatePackage/WingetCreatePackage.wapproj
```diff
@@ -1,70 +1,72 @@
...
+    <DefaultLanguage>en-US</DefaultLanguage>
+    <EntryPointProjectUniqueName>..\WingetCreateCLI\WingetCreateCLI.csproj</EntryPointProjectUniqueName>
+    <AppxAutoIncrementPackageRevision>True</AppxAutoIncrementPackageRevision>
+    <AppxBundlePlatforms>x86|x64</AppxBundlePlatforms>
+    <AppxBundle>Always</AppxBundle>
+    <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
+    <GenerateTestArtifacts>True</GenerateTestArtifacts>
+    <HoursBetweenUpdateChecks>0</HoursBetweenUpdateChecks>
+    <AppxPackageSigningTimestampDigestAlgorithm>SHA256</AppxPackageSigningTimestampDigestAlgorithm>
+    <AppxPackageDir>C:\win\tmp\wingetcreate-package\</AppxPackageDir>
```
yao-msft, line 41: line 40 and 41 were added accidentally

src/WingetCreatePackage/Package.appxmanifest
```diff
@@ -11,7 +11,7 @@
   <Identity
     Name="Microsoft.WindowsPackageManagerManifestCreator"
     Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"
-    Version="0.0.1.0" />
+    Version="0.0.2.0" />
```
yao-msft, line 14: nit: not needed but probably ok

src/WingetCreateCLI/Models/DscModels/SettingsResourceObject.cs
AmelBawa-msft: We don't have nullables enabled in the project, so the default for a string is `null`.
Since we don't have nullable reference enabled, using `string?` will cause a CS8632 warning.

yao-msft (approved):

PR #615, opened 2025-07-12
Added --allow-unsecure-downloads option for HTTP downloads

AmelBawa-msft:
By default only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.

### Help message
```
PS C:\> wingetcreate new --help
...
  --allow-unsecure-downloads    Allow unsecure downloads (HTTP) for this operation.
...
```
### Unsecure download
```
PS C:\> wingetcreate new http://mock
Downloading and parsing: http://mock...
Failed to download installer.
Only HTTPS URLs are supported without "--allow-unsecure-downloads"
```
### Unsupported protocol
```
PS C:\> wingetcreate new ftp://mock
Downloading and parsing: ftp://mock...
Failed to download installer.
Only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.
```

### DIFF
--- a/src/WingetCreateCLI/Commands/NewCommand.cs
+++ b/src/WingetCreateCLI/Commands/NewCommand.cs
@@ -75,6 +75,12 @@ namespace Microsoft.WingetCreateCLI.Commands
         public string OutputDir { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to allow unsecure downloads.
+        /// </summary>
+        [Option("allow-unsecure-downloads", Required = false, HelpText = "AllowUnsecureDownloads_HelpText", ResourceType = typeof(Resources))]
+        public bool AllowUnsecureDownloads { get; set; }
+
+        /// <summary>
         /// Gets or sets the format of the output manifest files.
         /// </summary>
         [Option('f', "format", Required = false, HelpText = "ManifestFormat_HelpText", ResourceType = typeof(Resources))]
@@ -116,7 +122,7 @@ namespace Microsoft.WingetCreateCLI.Commands
 
                 foreach (var installerUrl in this.InstallerUrls)
                 {
-                    string packageFile = await DownloadPackageFile(installerUrl);
+                    string packageFile = await DownloadPackageFile(installerUrl, this.AllowUnsecureDownloads);
                     if (string.IsNullOrEmpty(packageFile))
                     {
                         return false;
--- a/src/WingetCreateCLI/Properties/Resources.resx
+++ b/src/WingetCreateCLI/Properties/Resources.resx
@@ -1475,7 +1475,7 @@ Warning: Using this argument may result in the token being logged. Consider an a
     <value>Only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.</value>
   </data>
   <data name="DownloadHttpsOnly_Error" xml:space="preserve">
-    <value>Only HTTPS URLs are supported without "--allow-unsecure-downloads"</value>
+    <value>Only HTTPS URLs are supported without "--allow-unsecure-downloads".</value>
   </data>
   <data name="AllowUnsecureDownloads_HelpText" xml:space="preserve">
     <value>Allow unsecure downloads (HTTP) for this operation.</value>
--- a/src/WingetCreateCLI/Properties/Resources.resx
+++ b/src/WingetCreateCLI/Properties/Resources.resx
@@ -1472,7 +1472,7 @@ Warning: Using this argument may result in the token being logged. Consider an a
     <value>The scope value for Microsoft Entra Id authentication</value>
   </data>
   <data name="DownloadProtocolNotSupported_Error" xml:space="preserve">
-    <value>Only HTTP and HTTPS URLs are supported.</value>
+    <value>Only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.</value>
   </data>
   <data name="DownloadHttpsOnly_Error" xml:space="preserve">
     <value>Only HTTPS URLs are supported without "--allow-unsecure-downloads"</value>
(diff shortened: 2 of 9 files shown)


src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1471,4 +1471,13 @@ Warning: Using this argument may result in the token being logged. Consider an a
...
     <value>The scope value for Microsoft Entra Id authentication</value>
   </data>
+  <data name="DownloadProtocolNotSupported_Error" xml:space="preserve">
+    <value>Only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.</value>
+  </data>
+  <data name="DownloadHttpsOnly_Error" xml:space="preserve">
+    <value>Only HTTPS URLs are supported without "--allow-unsecure-downloads".</value>
+  </data>
+  <data name="AllowUnsecureDownloads_HelpText" xml:space="preserve">
+    <value>Allow unsecure downloads (HTTP) for this operation.</value>
```
mdanish-kh, line 1481:
```suggestion
    <value>Allow unsecure downloads (HTTP) for this operation. Default is false.</value>
```

For consistency with other help texts of boolean params

image

src/WingetCreateCLI/Properties/Resources.resx
```diff
@@ -1471,4 +1471,13 @@ Warning: Using this argument may result in the token being logged. Consider an a
   <data name="MicrosoftEntraIdAuthenticationInfo_Scope_KeywordDescription" xml:space="preserve">
     <value>The scope value for Microsoft Entra Id authentication</value>
   </data>
+  <data name="DownloadProtocolNotSupported_Error" xml:space="preserve">
+    <value>Only HTTPS URL are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.</value>
```
mdanish-kh, line 1475:
```suggestion
    <value>Only HTTPS URLs are supported for downloads. Use the "--allow-unsecure-downloads" option to allow HTTP URLs.</value>
```

nit: grammar & since you've used URL`s` everywhere else

src/WingetCreateCLI/Commands/NewCommand.cs
```diff
@@ -74,6 +74,12 @@ public static IEnumerable<Example> Examples
         [Option('o', "out", Required = false, HelpText = "OutputDirectory_HelpText", ResourceType = typeof(Resources))]
         public string OutputDir { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to allow unsecure downloads.
+        /// </summary>
+        [Option("allow-unsecure-downloads", Required = false, HelpText = "AllowUnsecureDownloads_HelpText", ResourceType = typeof(Resources))]
+        public bool AllowUnsecureDownloads { get; set; }
```
mdanish-kh, line 81: We should update the arguments section of [doc/new.md](https://github.com/microsoft/winget-create/blob/main/doc/new.md) & [doc/update.md](https://github.com/microsoft/winget-create/blob/main/doc/update.md) as well

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -1116,5 +1119,23 @@ private static string RemoveInvalidCharsFromString(string value)
         {
             return Regex.Replace(value, InvalidCharacters, string.Empty);
         }
+
+        private static void ValidateUrl(string url, bool allowHttp)
```
mdanish-kh, line 1123: Since you've added an explicit function for checking if a string is a valid URI for our needs, we _may_ want to change this code in UpdateCommand.cs (that checks if the param `--release-notes-url` is a valid URL) to re-use this function for consistency

https://github.com/microsoft/winget-create/blob/c92436b01095e45b054652116b7ff0eef28c7ca0/src/WingetCreateCLI/Commands/UpdateCommand.cs#L170-L178

AmelBawa-msft: This change specifically targets scenarios where bits are downloaded. We retrieve the installer from the provided input URL to parse metadata such as the publisher, version, and other relevant details. To enhance security, we want to ensure that both the download and the parsing of the package occur over a secure protocol, such as HTTPS.

src/WingetCreateCLI/Properties/Resources.resx
AmelBawa-msft: Agree this does fall into a bit of a gray area 😉 That said, I think it's reasonable to assume that if the option isn't explicitly specified, the feature should be considered disabled by default.

yao-msft:
I feel the security team will not be happy with this change. Is there a reason we allow http downloads? I remember winget only accepts https.

yao-msft:
Ok, digging through my very ancient email threads. now I remember we "tried" to limit to https only but it affects a good portion of the packages in the repo. And since we have sha256 verification, we end up not doing the restriction.

src/WingetCreateCore/Common/PackageParser.cs
```diff
@@ -117,16 +117,19 @@ public static void ParsePackages(List<InstallerMetadata> installerMetadataList,
...
+        public static async Task<string> DownloadFileAsync(string url, bool allowHttp, long? maxDownloadSize = null)
         {
+            ValidateUrl(url, allowHttp);
             var response = await httpClient.GetAsync(url, HttpCompletionOption.ResponseHeadersRead);
 
             int redirectCount = 0;
             while (response.StatusCode == System.Net.HttpStatusCode.Redirect && redirectCount < 2)
             {
                 var redirectUri = response.Headers.Location;
+                ValidateUrl(url, allowHttp);
```
JohnMcPMS, line 132:
```suggestion
                ValidateUrl(redirectUri, allowHttp);
```
Presumably you don't feel a need to check the same URL again.

yao-msft (approved):
I think a full fix (from security perspective) would be make this switch behind something like admin settings/group policy. But I don't know if we'll want to go that far. Anyway this is a good first step towards that.

Issue #517, opened 2024-02-20
Add support for removing manifests


vedantmgoyal9:
### Description of the new feature / enhancement

Add remove command to winget-create.

denelon:
We're certainly looking at a Microsoft.WinGet.Source module for managing WinGet sources as well.

PR #625, opened 2025-09-24
Add workflow for automatic issue deduplication

cinnamon-msft:

-----

This adds a bot to automatically detect duplicate issues after an issue is opened or re-opened. It can only apply the "duplicate" label. It will not close the issue and it will leave a comment with the duplicate issue(s).


AmelBawa-msft:
Example of similar-issues-ai app referenced by @vedantmgoyal9: https://github.com/microsoft/winget-create/issues/517#issuecomment-1954833004

PR #627, opened 2025-09-25
Support specifying the fork owner

martincostello:
Allow specifying the owner for a fork, instead of assuming the GitHub token is associated with a user, to allow organizations to fork a repository and have a GitHub app installed into that repository to submit pull requests.

The use case here is that we want to fork [microsoft/winget-pkg](https://github.com/microsoft/winget-pkgs) to `grafana/winget-pkg`, and then use a GitHub app installed into that fork to submit pull requests for new manifests.

-----

I haven't yet manually verified this works, and haven't yet added unit tests. Just floating now to see if this is functionality that would be accepted.


martincostello:
From manual testing, it seems like this fundamentally doesn't work with the GitHub API using an app that isn't installed into the target repository. If you're in a fork from one organization with its own GitHub app, you don't seem to be able to submit a pull request from that organization's fork back to the microsoft origin/upstream. 😞

PR #688, opened 2026-07-29
Update pipeline configuration

AmelBawa-msft:
## 📖 Description
Update pipeline configuration

## ✅ Checklist

- [ ] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue
- [ ] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-create/pull/688)

### DIFF
--- a/pipelines/azure-pipelines.release.yml
+++ b/pipelines/azure-pipelines.release.yml
@@ -175,7 +175,7 @@ extends:
               inputs:
                 command: restore
                 feedsToUse: config
-                nugetConfigPath: "NuGet.Internal.config"
+                nugetConfigPath: "nuget.config"
                 projects: $(workingDirectory)/**/*.csproj
                 arguments: --runtime win-x64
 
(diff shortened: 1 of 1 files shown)


pipelines/azure-pipelines.release.yml
```diff
@@ -34,9 +34,13 @@ extends:
   parameters:
     globalSdl:
       tsa:
-        enabled: false
+        enabled: true
         configFile: $(Build.SourcesDirectory)\.config\tsaoptions.json
-        onboard: false
+        onboard: true
+      armory:
+        enabled: false
```
florelis, line 41: Was this causing issues? I don't think we even have any ARM templates for it to complain about, do we?

florelis (approved):

pipelines/azure-pipelines.release.yml
AmelBawa-msft: You're right we have no ARM templates, so Armory has nothing to scan, but it still tries to upload its (empty) results to TSA, and that upload call itself fails. Disabling  `globalSdl.armory.enabled` skips it entirely with no coverage loss, since it's a real flag the 1ES template respects

